PR #13, opened 2026-05-13
fix(terminal): always scroll buffer on wheel, regardless of mouse tracking

yong076:
## Summary

Inside a wmux pane, TUIs that enable xterm mouse tracking via DECSET 1000/1002/1003/1006 (Claude Code, vim, tmux, lazygit, htop, …) broke wheel scrolling and painted a small highlight rectangle that followed the mouse cursor. This PR fixes both with a single capture-phase wheel listener.

## Root cause

xterm.js's wheel handler forwards wheel events to the application instead of scrolling its own buffer when `requestedEvents.wheel` is set (see `@xterm/xterm` `Terminal.ts`: `if (requestedEvents.wheel) return`). Two visible symptoms share that root cause:

1. **Scrollback dead** — wheel events get encoded as button-4/5 mouse reports and sent to the app; xterm's buffer never moves.
2. **Rectangle tracking the cursor** — the cell-resolution motion reports the app receives weren't being consumed for scroll, so the app rendered them as a hover highlight.

## Fix

Install a capture-phase `wheel` listener on the terminal host element. It `preventDefault`s the event and calls `terminal.scrollLines(...)` directly, so xterm's own listener never runs and the app never sees the wheel event.

What's preserved:
- Mouse-button reports still work (clicks, drags, button release)
- Mouse-motion reports still work (just not as wheel)
- Selection, copy, paste behavior unchanged
- WebGL renderer unchanged

## Test plan

- [x] Type-check passes (`npm run build:main`)
- [x] Unit tests pass (5 pre-existing `pty-manager.test.ts` failures unrelated to this change; identical on `master`)
- [ ] Manual: open Claude Code / vim / tmux inside a pane, wheel-scroll works, no hover rectangle

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

### DIFF
--- a/src/renderer/hooks/useTerminal.ts
+++ b/src/renderer/hooks/useTerminal.ts
@@ -192,6 +192,35 @@ export function useTerminal({ surfaceId, shell, cwd, visible = true, focused = t
     // Open terminal in the DOM
     terminal.open(terminalRef.current);
 
+    // Always scroll wmux's buffer on wheel — never forward to the app.
+    // Without this, when a TUI (Claude Code, vim, tmux…) enables mouse
+    // tracking via DECSET 1000/1002/1003/1006, xterm.js sends wheel
+    // events to the app instead of scrolling the buffer (see
+    // @xterm/xterm Terminal.ts wheel handler: `if (requestedEvents.wheel) return`).
+    // That has two visible effects: scrollback is dead, AND the app paints
+    // a cell highlight that tracks the mouse cursor. We intercept on the
+    // capture phase before xterm's listener runs.
+    const wheelHost = terminalRef.current;
+    const onWheelCapture = (ev: WheelEvent) => {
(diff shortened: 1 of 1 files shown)


amirlehmam:
Thank you @yong076 — merged and shipped in **v0.7.20**.

The root-cause write-up was very useful: pinning the symptom (hover rectangle + dead scrollback) onto a single line in xterm's wheel handler (`if (requestedEvents.wheel) return`) made the fix obvious. The capture-phase listener is the right minimal intervention — no patching of xterm internals, no behavioral change for apps that don't enable mouse tracking.

Manually verified inside Claude Code and lazygit panes: wheel scrolls the buffer, no ghost rectangle, mouse clicks/selection still work.

Released: https://github.com/amirlehmam/wmux/releases/tag/v0.7.20

PR #14, opened 2026-05-13
feat: WMUX_INSTANCE for side-by-side dev/prod instances

yong076:
## Summary

Adds a `WMUX_INSTANCE=<name>` environment variable that lets a dev build of wmux run alongside an installed production wmux without colliding on the named pipe (which is exclusive on Windows) or overwriting the user's `session.json`.

Set `WMUX_INSTANCE=dev` and the running instance uses:
- Pipe: `\.\pipe\wmux-dev` instead of `\.\pipe\wmux`
- APPDATA: `%APPDATA%\wmux-dev\` instead of `%APPDATA%\wmux\`

Unset → unchanged default behavior, no observable difference for existing users.

## Motivation

Today, running `npm run dev` while a daily-driver wmux is running fights over:
1. **The named pipe** — `PipeServer.start()` fails to bind because the production instance already owns `\.\pipe\wmux`. CLI/agent commands from the dev instance break silently.
2. **`%APPDATA%\wmux\sessions\session.json`** — both instances auto-save (30s), so the dev instance overwrites the user's live workspace state. On next restart of the daily-driver wmux, sessions are gone.

This makes verifying fixes in dev hostile to the user's actual work. With this PR you can:

```powershell
$env:WMUX_INSTANCE = "dev"
npm run dev
```

…and the dev instance is fully isolated.

## Design

One tiny shared helper (`src/shared/instance.ts`) exposes `getPipePath()` and `getAppDataDir()`. Both derive a suffix from `process.env.WMUX_INSTANCE`. Five sites now go through the helper:

- `src/main/index.ts` — `new PipeServer(getPipePath())`
- `src/main/pty-manager.ts` — `WMUX_PIPE` injected into PTY env
- `src/main/session-persistence.ts` — `APPDATA_DIR` source
- `src/cli/wmux.ts`, `src/cli/wmux-hook.ts` — prefer `process.env.WMUX_PIPE` (which the main process injects), fall back to the default

That last bit is what makes the CLI inside a spawned shell talk back to the correct instance.

## Test plan

- [x] `npm run build:main` passes
- [x] `pipe-server.test.ts` (4) and `session-persistence.test.ts` (4) pass
- [x] No public API changes; default behavior unchanged when `WMUX_INSTANCE` is unset
- [ ] Manual: run two instances side-by-side, confirm pipes and APPDATA are independent

## Notes

- The 5 failing `pty-manager.test.ts` tests on this branch are pre-existing and identical on `master` (Windows shell-spawn environment issue).
- The unrelated wheel-scroll fix lives in a separate PR.

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

### DIFF
--- a/src/cli/wmux-hook.ts
+++ b/src/cli/wmux-hook.ts
@@ -7,7 +7,7 @@
 import net from 'net';
 
 const tool = process.argv[2] || 'unknown';
-const pipePath = '\\\\.\\pipe\\wmux';
+const pipePath = process.env.WMUX_PIPE || '\\\\.\\pipe\\wmux';
 
 const client = net.connect({ path: pipePath }, () => {
   const msg = JSON.stringify({ method: 'hook.event', params: { tool }, id: 1 });
--- a/src/cli/wmux.ts
+++ b/src/cli/wmux.ts
@@ -2,7 +2,9 @@
 
 import net from 'net';
 
-const PIPE_PATH = '\\\\.\\pipe\\wmux';
+// Respect WMUX_PIPE when set (e.g. by a parent wmux running with WMUX_INSTANCE),
+// so the CLI talks to the same instance that spawned the shell.
+const PIPE_PATH = process.env.WMUX_PIPE || '\\\\.\\pipe\\wmux';
 
 function sendV1(command: string): Promise<string> {
   return new Promise((resolve, reject) => {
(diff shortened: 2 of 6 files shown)


amirlehmam:
Thank you @yong076 — merged and shipped in **v0.7.20**.

This is exactly the kind of clean isolation I'd been wanting. Two details I appreciated:

1. The CLI fallback through `process.env.WMUX_PIPE` (which the main process injects into spawned shell env) means the CLI inside a dev-instance pane talks back to that same dev instance — no manual wiring needed.
2. Default behavior is unchanged when `WMUX_INSTANCE` is unset, so this is risk-free for existing users.

Documented in the v0.7.20 release notes. Going forward I'll use `WMUX_INSTANCE=dev` for development so the daily-driver wmux is safe from `session.json` clobbering.

Released: https://github.com/amirlehmam/wmux/releases/tag/v0.7.20

PR #24, opened 2026-06-06
fix(browser): remove hardcoded 800px max width on browser pane

schroldgames:
## Summary

- Replaces the hardcoded `800px` upper bound on browser pane width with `window.innerWidth - 400`, so the pane can scale with the window
- On larger displays (or when the terminal side only needs a narrow strip) the browser pane was artificially capped at 800px regardless of available space
- Also adds `.claude/settings.local.json` to `.gitignore` (local Claude Code settings should not be committed)

## Behaviour

| Before | After |
|--------|-------|
| Drag handle stops at 800px even on wide displays | Drag handle goes up to `windowWidth - 400px`, leaving at least 400px for the terminal side |

## Test plan

- [ ] Open a browser pane and drag the resize handle — confirm it can expand beyond 800px on a display wider than ~1200px
- [ ] Confirm the terminal side never collapses below 400px
- [ ] Confirm minimum browser pane width (250px) is still respected when dragging toward the terminal

### DIFF
--- a/.claude/settings.local.json
+++ b/.claude/settings.local.json
@@ -1,7 +1,0 @@
-{
-  "permissions": {
-    "allow": [
-      "Bash(npx tsc:*)"
-    ]
-  }
-}
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,6 @@ release/
 
 # Local Netlify folder
 .netlify
+
+# Local Claude Code settings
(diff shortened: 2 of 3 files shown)


amirlehmam:
Merged and shipped in **v0.8.3**. Thanks @schroldgames — the browser pane now scales with the window instead of capping at 800px.

PR #25, opened 2026-06-06
fix(browser): scope navigate events by surfaceId

schroldgames:
## Summary

- All `wmux:browser-navigate` CustomEvents were broadcasts — every open `BrowserPane` responded, so the wrong pane would navigate when multiple browser surfaces existed, and dev-server auto-navigate fired on all of them indiscriminately
- Adds a `surfaceId` field to the event detail; `BrowserPane` ignores events that don't target its own id (events with no `surfaceId` remain broadcasts for back-compat)
- `open-in-browser.ts` resolves the target surface's id from the split tree and passes it in both the immediate and deferred navigate dispatches
- `App.tsx` auto-navigate on dev-server port detection now passes the workspace's browser surface id instead of an empty string

## Files changed

| File | Change |
|------|--------|
| `src/preload/index.ts` | Forward `surfaceId` arg through the navigate event detail |
| `src/renderer/components/Browser/BrowserPane.tsx` | Scope event handler — ignore events targeting a different surface |
| `src/renderer/utils/open-in-browser.ts` | Resolve browser surfaceId from split tree; pass it in navigate dispatches |
| `src/renderer/App.tsx` | Pass correct surfaceId on dev-server auto-navigate |

## Test plan

- [ ] Open two workspaces each with a browser pane — navigating one should not affect the other
- [ ] Trigger dev-server auto-navigate (start a local server on a detected port) — confirm only the active workspace's browser pane navigates
- [ ] `wmux browser open <url>` from CLI — confirm it targets the correct pane
- [ ] Backwards compat: dispatch `wmux:browser-navigate` with no `surfaceId` — confirm it still navigates the first available pane

### DIFF
--- a/.claude/settings.local.json
+++ b/.claude/settings.local.json
@@ -1,7 +1,0 @@
-{
-  "permissions": {
-    "allow": [
-      "Bash(npx tsc:*)"
-    ]
-  }
-}
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,6 @@ release/
 
 # Local Netlify folder
 .netlify
+
+# Local Claude Code settings
+.claude/settings.local.json
--- a/src/preload/index.ts
+++ b/src/preload/index.ts
@@ -65,9 +65,9 @@ contextBridge.exposeInMainWorld('wmux', {
     },
   },
   browser: {
-    navigate: (_surfaceId: string, url: string) => {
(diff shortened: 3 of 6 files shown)


amirlehmam:
Merged and shipped in **v0.8.3**. Thanks @schroldgames — surface-scoped navigate events pair nicely with the new browser/markdown tab menu (#28).

PR #37, opened 2026-06-17
fix(terminal): clean profile/terminal startup under oh-my-posh

### LINKED ISSUE #36: Terminal/profile startup glitches with oh-my-posh: DA escape leak, doubled prompt, slow first prompt
## Summary

Opening a terminal in wmux — most visibly via a **quick-launch profile** — produces a broken first prompt when the shell uses **oh-my-posh** (PowerShell + PSReadLine). Three symptoms, all rooted in how terminal **queries/responses** and **sizing** are handled across wmux's multi-process (renderer ⇄ main ⇄ pty) architecture.

## Environment

- Shell: PowerShell (`pwsh`) with **oh-my-posh** prompt + PSReadLine
- wmux terminal: xterm.js (renderer) ⇄ Electron main ⇄ node-pty/ConPTY
- Repro: open a quick-launch profile (even one that only sets a `cwd`), or any new terminal tab

## Symptoms

1. **Garbled characters on the prompt.** A stray escape sequence appears as if typed:
   ```
   ╰─❯ [?62;4;9;22c
   ```
   Typing `cls` to clear it yields `62;4;9;22ccls` — the junk and the typed command merge into one executed line.

2. **Doubled prompt.** The oh-my-posh prompt renders twice on startup.

3. **Slow / stalled first prompt.** The terminal takes several seconds before the prompt appears.

## Root causes

- **DA1 (Device Attributes) response leak.** oh-my-posh / PSReadLine probe the terminal with a DA1 query (`ESC [ c`). The xterm **image addon** answers it with `ESC [ ? 62 ; 4 ; 9 ; 22 c` (62=VT220, 4=Sixel, 9, 22=ANSI color). That reply travels a slow renderer→main→pty round-trip and arrives **after** the prompt is drawn, so PSReadLine echoes it as typed input → the `[?62;4;9;22c` junk.

- **Doubled prompt = two shells per tab.** React **StrictMode** double-mounts the terminal component in dev; the renderer's `pty.has()` check is async, so `pty.create()` could fire **twice** for one `surfaceId`, spawning a **second PowerShell process** that also streamed to the renderer (and leaked an orphan). Separately, spawning the PTY at the default 80×24 and then resizing it made the shell **redraw** the prompt.

- **Slow first prompt.** The shell-integration script started the PR-polling background job with `Start-Job` **during init** — `Start-Job` spins up a child PowerShell runspace and costs several hundred ms on the startup critical path.

## Impact

Every new terminal/profile under oh-my-posh starts dirty: an executed garbage command, a duplicated prompt, and a multi-second delay. Diagnosed with an opt-in PTY I/O trace that showed the duplicate shell-init sequences and the DA round-trip timing.

## Proposed fix

- Answer DA1 **in-process** in `PtyManager` (instant) and suppress xterm's own (slow) DA1 reply.
- Make `PtyManager.create()` **idempotent per `surfaceId`**; skip re-wiring a reused PTY in the IPC handler.
- Spawn PTYs at the **measured size**; drop redundant **same-size resizes**.
- **Defer** the PR `Start-Job` to the shell's first idle.
- Bake quick-launch startup commands into shell init for PowerShell instead of injecting keystrokes.
- Guard async terminal callbacks against use-after-dispose (fixes an xterm `reading 'dimensions'` crash).

kevmtt:
Fixes broken terminal/profile startup under oh-my-posh (PowerShell + PSReadLine): garbled prompt characters, a doubled prompt, and a slow first prompt.

Closes #36

## Root causes & fixes

| Symptom | Root cause | Fix |
|---------|-----------|-----|
| `[?62;4;9;22c` junk echoed on the prompt | xterm image addon's **DA1 reply** does a slow renderer↔main round-trip and lands after the prompt → PSReadLine echoes it as input | Answer DA1 **in-process** in `PtyManager` (instant); suppress xterm's own DA1 reply in `useTerminal` so there's no slow duplicate |
| Doubled prompt | React **StrictMode** double-mount + async `pty.has()` → `pty.create()` fired twice → **two shells** under one `surfaceId` | `PtyManager.create()` is **idempotent per surfaceId**; IPC handler skips re-wiring a reused PTY |
| Doubled prompt (2nd cause) | PTY spawned at 80×24 then resized → shell redraws prompt | Spawn PTY at the **measured size**; drop redundant **same-size resizes** |
| Slow first prompt | PR-polling `Start-Job` (spins a child runspace) ran during shell init | **Defer** it to the shell's first idle |
| Quick-launch startup commands racing init queries | startup commands injected as keystrokes after a fixed delay | **Bake** them into shell init for PowerShell |
| xterm `reading 'dimensions'` crash | async callbacks ran after `terminal.dispose()` (StrictMode) | Guard callbacks with a `disposed` flag + cancel pending rAFs |

## Files

- `src/main/pty-manager.ts` — in-process DA1 reply, idempotent create, spawn-at-size, same-size resize drop
- `src/main/ipc-handlers.ts` — skip re-wiring reused PTYs
- `src/preload/index.ts` — `cols`/`rows` in create options
- `src/renderer/hooks/useTerminal.ts` — DA1 suppression, dispose guards, spawn-at-measured-size
- `src/shell-integration/wmux-powershell-integration.ps1` — deferred `Start-Job`, baked startup commands

## Test plan

- [x] `npm run build:main` + `tsc` (main & renderer) clean; `eslint` 0 errors on changed files
- [x] Open a quick-launch profile under oh-my-posh → **single clean prompt**, no `[?62;4;9;22c`, no stall (verified via PTY trace: one shell-init sequence, one prompt render)
- [x] DA1 answered in-process (trace showed instant reply)
- [ ] Sanity: normal tabs, split panes, tab switching, agent-spawned terminals still attach correctly

## Notes

All temporary diagnostic instrumentation has been removed; only the fixes remain, each with an explanatory comment.

### DIFF
--- a/src/main/ipc-handlers.ts
+++ b/src/main/ipc-handlers.ts
@@ -44,6 +44,12 @@ export function registerIpcHandlers(windowManager: WindowManager, cdpProxyInstan
       };
       const created = ptyManager.create(resolvedOptions);
       const id = created.id;
+      // Reused PTY (idempotent create — e.g. StrictMode's double create() race):
+      // the original create already wired data/exit forwarding. Re-wiring here
+      // would forward every chunk twice and double everything in the renderer.
+      if (created.reused) {
+        return created;
+      }
       const window = BrowserWindow.fromWebContents(_event.sender);
       const unsubData = ptyManager.onData(id, (data) => {
         if (window && !window.isDestroyed()) {
--- a/src/main/pty-manager.ts
+++ b/src/main/pty-manager.ts
@@ -95,6 +95,15 @@ interface PtyEntry {
   writeChain: Promise<void>;
   pendingChunks: number;
   alive: boolean;
+  // Last applied size. Used to drop redundant same-size resizes, which would
+  // otherwise make the shell (PSReadLine/oh-my-posh) redraw the prompt for no
+  // reason — a cause of the doubled prompt on startup.
+  cols: number;
+  rows: number;
+  // Resolved shell + whether startup commands were baked in — returned verbatim
+  // when create() is called again for the same surfaceId (idempotent reuse).
+  shell: string;
+  startupConsumed: boolean;
 }
 
 export interface CreateOptions {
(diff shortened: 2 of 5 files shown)


kevmtt:
I'm honest here.. Many things I don't understood here, so I let AI add comments for me to understand. I left them in case they might be usefull.

amirlehmam:
Merged and shipped in **v0.8.8** 🚀 Thanks @kevmtt! The in-process DA1 reply + idempotent `PtyManager.create()` + spawn-at-measured-size all landed cleanly and compose well with the security PR (#45) that also touched `pty-manager.ts`. Clean single prompt under oh-my-posh confirmed.

PR #42, opened 2026-06-18
fix(terminal): make scroll wheel work again in alternate screen

### LINKED ISSUE #41: Mouse wheel scroll broken inside Claude Code (alternate screen buffer) — v0.8.7
## Environment
- wmux: v0.8.7
- Shell: WSL running Ubuntu

## Behavior
Mouse wheel scroll does not work inside Claude Code (or any full-screen TUI on the alternate screen buffer).

Scroll works fine in normal shell panes. PgUp/PgDn works as a workaround.

## Expected behavior
Mouse wheel scrolls the terminal output, same as in normal panes.

## Notes
This was explicitly fixed previously — the v0.7.21 release notes mention the same bug and describe the fix (only intercept wheel on normal buffer, pass through on alternate screen). Appears to have regressed in v0.8.x.

egertaia:
Fixes #41 

## Summary

- Fixes scroll wheel regression in alternate-screen apps (Claude Code, vim, less, htop) introduced in v0.8.4+
- When on the alternate buffer, takes ownership of wheel events and forwards them to the PTY directly instead of relying on xterm's `_handleWheel` / `_handlePassiveWheel` which are unreliable after the WebGL context swap
- Sends SGR mouse-wheel sequences (`\x1b[<64;col;rowM` / `\x1b[<65;col;rowM`) when the app has enabled mouse tracking, falls back to arrow keys otherwise

## Root cause

xterm's built-in wheel forwarding fails in two cases:
1. The wheel listener is only registered when the app has explicitly requested wheel mouse events via DECSET
2. `_consumeWheelEvent` returns 0 when render-service cell dimensions are temporarily unavailable (can happen after the WebGL renderer swap in v0.8.4+), causing the event to be silently dropped

## Test plan

- [ ] Scroll in a normal terminal — scrollback works as before
- [ ] Scroll inside Claude Code / vim / less — content scrolls correctly
- [ ] Scroll in a TUI without mouse tracking enabled — arrow keys sent, content moves

### DIFF
--- a/src/renderer/hooks/useTerminal.ts
+++ b/src/renderer/hooks/useTerminal.ts
@@ -210,15 +210,71 @@ export function useTerminal({ surfaceId, shell, cwd, visible = true, focused = t
     const wheelHost = terminalRef.current;
     const onWheelCapture = (ev: WheelEvent) => {
       if (ev.deltaY === 0) return;
-      // Only hijack the wheel for scrollback on the NORMAL buffer. Full-screen
-      // TUIs (Claude Code, vim, less, htop…) switch to the ALTERNATE buffer
-      // (DECSET 1049), which has no scrollback — terminal.scrollLines() there is
-      // a no-op. If we still preventDefault/stopPropagation we also suppress
-      // xterm's native behavior of forwarding the wheel to the app (mouse-wheel
-      // reports when mouse tracking is on, otherwise arrow-key sequences), which
-      // is the ONLY way those apps scroll their own content. So on the alt
-      // buffer we fall through and let xterm handle it.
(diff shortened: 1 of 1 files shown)


amirlehmam:
Merged and shipped in **v0.8.8** 🚀 Thanks @kevmtt! Your alt-buffer ownership fix (#41) is the foundation of the unified wheel handler — I merged it together with #48 (the webview/tmux case). Since the handler now always takes ownership on the capture phase, the SGR-vs-arrow-keys decision uses #48's remount-safe `surfaceMouseEnabled` signal while keeping your arrow-key fallback for non-mouse pagers. See the note on #48 for the full resolution.

PR #44, opened 2026-06-18
fix(clipboard): correct UTF-8 for terminal copy/paste (OSC 52)

egertaia:
## Problem

Copying text containing multi-byte UTF-8 characters from a TUI app (Claude Code, or tmux with `set-clipboard on`) and pasting it back produced mojibake — e.g. an em dash `—` became `â€"`.

Root cause: these apps copy via the **OSC 52** clipboard escape sequence, whose payload is base64-encoded UTF-8. The handler decoded it with `atob()`, which returns a binary/Latin-1 string (one code point per byte), so the em dash bytes `E2 80 94` turned into code points `U+00E2 U+0080 U+0094`.

Plain xterm text-selection copy goes through `navigator.clipboard.writeText()` and was unaffected — which is exactly why "copy from the terminal works, copy from Claude doesn't."

## Fix

- **OSC 52 decode**: decode the base64 payload as UTF-8 (`TextDecoder` over the raw bytes) instead of `atob()`'s Latin-1 string. *(the actual bug)*
- **Paste hardening** (consistency):
  - Read the clipboard via Electron's `clipboard.readText()` instead of `navigator.clipboard.readText()` (no focus/permission requirement).
  - Route the `Ctrl+Shift+V` paste action through the terminal so it honors bracketed-paste mode — multi-line paste no longer submits on the first newline (e.g. pasting into Claude Code).

## Verification

Reproduced with an instrumented build logging clipboard reads/writes:
- Before: copying `foo — bar` from a TUI logged `U+00E2 U+0080 U+0094`.
- After: same copy logs `U+2014`, and the paste round-trips as `foo — bar`.

### DIFF
--- a/src/main/ipc-handlers.ts
+++ b/src/main/ipc-handlers.ts
@@ -202,6 +202,10 @@ export function registerIpcHandlers(windowManager: WindowManager, cdpProxyInstan
     clipboard.writeText(text);
   });
 
+  // Use Electron's clipboard for reads too — navigator.clipboard.readText() can
+  // return garbled text on Windows when the source app wrote a non-UTF-8 format.
+  ipcMain.handle('clipboard:read-text', () => clipboard.readText());
+
   // Clipboard image paste: save clipboard image to temp file, return path
   ipcMain.handle('clipboard:paste-image', async () => {
     const img = clipboard.readImage();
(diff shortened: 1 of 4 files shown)


amirlehmam:
Merged and shipped in **v0.8.8** 🚀 Thanks @egertaia! The OSC 52 UTF-8 decode (TextDecoder over the raw base64 bytes instead of `atob`) auto-merged cleanly against the terminal refactor. The bracketed-paste routing for Ctrl+Shift+V is a nice consistency win too.

Issue #9, opened 2026-05-09
Size not fixed when clicking settings tab


whrhkdrhs-sketch:
I have decided to become a user to contribute to the endless development of wmux.

The first issue is that the window header does not stay fixed and moves up and down whenever I switch tabs in the settings window. I hope this can be fixed, whether by using a scrollbar or by referencing the settings UI in Visual Studio.

Thank you.

amirlehmam:
Hi @whrhkdrhs-sketch — thanks a lot for the report and welcome aboard! 👋

You were right on both counts: the settings window had no fixed height, so the overlay re-centered it whenever a tab swapped to taller/shorter content, which made the header look like it was jumping. The fix pins the window to a stable size and lets the inner content scroll instead — the same approach VS Code uses.

Patch: 70de573 (`src/renderer/styles/settings.css`).

Shipping in **v0.7.17**, building now: https://github.com/amirlehmam/wmux/actions/runs/25609656853 — once the workflow finishes the signed zip will be on the release page: https://github.com/amirlehmam/wmux/releases/tag/v0.7.17

Closing once the release goes live. Please keep the issues coming! 🙏

whrhkdrhs-sketch:
From a Windows user's perspective, wmux is really great. I will actively promote it. Thank you.

PR #3, opened 2026-04-15
feat: show shell-specific labels for terminal tabs

Optuber01:
## Summary
- return the resolved shell from PTY creation
- store the resolved shell on terminal surfaces
- show a shell-specific tab label instead of the generic Terminal label

## Why
When multiple terminals are open, shell-specific labels make it much easier to understand which tab is PowerShell, Command Prompt, Bash, WSL, or another shell.

## Testing
- npm run build:main
- npm run build:renderer

PR #2, opened 2026-04-15
feat: improve terminal and markdown link behavior

Optuber01:
## Summary
- open terminal and markdown links in the wmux browser when available
- keep Ctrl+Click and Cmd+Click opening the default system browser
- add clearer hover hints for terminal and markdown links

## Why
This makes link behavior more predictable inside wmux while still preserving an easy path to the external browser.

## Testing
- npm run build:main
- npm run build:renderer

Issue #1, opened 2026-04-09
Korean IME (한글) input issues on Windows 11


babyfings:
## Environment
- wmux version: 0.6.1
- OS: Windows 11 Pro (10.0.26200)
- IME: Windows default Korean IME (Microsoft 입력기)
- Previously tried third-party Korean IME — same issue

## Description
Korean (Hangul) composition input is unreliable. Characters sometimes get dropped, duplicated, or composed incorrectly during typing.

## Root Cause Analysis
After inspecting the source (`app.asar`), wmux has no Korean-specific IME handling. It relies entirely on xterm.js (v5.5.0) CompositionHelper.

Potential issues found:
1. `_finalizeComposition()` uses `setTimeout(0)` which may cause race conditions during Hangul composition
2. conpty PTY layer may have character encoding issues with multi-byte Korean characters
3. Electron's textarea composition events may not sync properly with Windows 11 IME

## Steps to Reproduce
1. Open wmux terminal on Windows 11
2. Switch to Korean IME
3. Type Korean characters — observe irregular composition behavior

## Expected
Smooth Korean character composition, same as Windows Terminal or VS Code integrated terminal.

amirlehmam:
Thanks for the excellent bug report — your root-cause analysis was spot on. Fixed in **[v0.6.3](https://github.com/amirlehmam/wmux/releases/tag/v0.6.3)** (commit 5292107).

## Root cause

Your point #1 was the actual bug. In `CompositionHelper._finalizeComposition()` (xterm.js 5.5, `src/browser/input/CompositionHelper.ts`), the `compositionend` path defers reading the textarea via `setTimeout(0)`:

```ts
this._isSendingComposition = true;
setTimeout(() => {
  if (this._isSendingComposition) {
    // ... reads this._textarea.value.substring(currentCompositionPosition.start, ...)
    this._coreService.triggerDataEvent(input, true);
  }
}, 0);
```

With fast Hangul typing, a new `compositionstart` can fire before the deferred reader runs, which shifts `_isComposing` / `_compositionPosition` so the substring that finally reaches `triggerDataEvent` is wrong — manifesting as the dropped / duplicated / incorrectly composed characters you saw. xterm's own inline comment on that block literally names Korean as the counter-example:

> *The last compositionupdate event's data property does not always accurately describe the character, a counter example being Korean where an ending consonant can move to the following character if the following input is a vowel.*

[...]

PR #45, opened 2026-06-19
security: harden pipe auth, CDP proxy, markdown XSS, and webview

foschmitz:
## Summary

Addresses the critical/high findings from a security review of the IPC pipe, CDP proxy, markdown rendering, and webview surfaces. No behavior change for normal use; the only externally-visible effect is that privileged pipe (V2) calls and direct CDP WebSocket connections from web origins are now rejected.

### #1 — Unauthenticated local RCE via named pipe (Critical)
Any local process could call privileged V2 methods (`agent.spawn`, `browser.eval`, `markdown.load_file`, workspace/pane mutation). Now:
- Per-instance token generated with `crypto.randomBytes(32)`, persisted `0600` under the instance APPDATA dir, compared with `crypto.timingSafeEqual`.
- Privileged V2 methods require the token; an **allowlist** of read-only/telemetry methods (`system.identify`, `system.capabilities`, `hook.event`, `agent.activity`) stays open so detection, shell hooks, and agent telemetry keep working — and any *new* method is locked by default.
- Token injected into spawned shells as `WMUX_PIPE_TOKEN`; the `wmux` CLI and hook client read it from env or the token file.
- V1 plaintext telemetry (shell integration) is unchanged (low-risk, no code-exec surface).

### #2 — CDP proxy reachable from web origins (Critical)
The CDP proxy exposes `Runtime.evaluate` (arbitrary JS in the webview). Loopback binding alone is insufficient:
- **DNS-rebinding**: HTTP endpoints + WS upgrade now require a loopback-literal `Host` (403 / `verifyClient` reject).
- **Direct WS from a browser page**: WebSockets bypass CORS, so a malicious page could open `ws://127.0.0.1:9222` (valid `Host`) and drive the debugger. We now also validate `Origin` on the WS upgrade and reject any web/file origin. Absent `Origin` is allowed so `chrome-devtools-mcp`/`puppeteer-core`/native `ws` clients still connect (mirrors Chrome's `--remote-allow-origins`).

### #3 — Markdown XSS (High)
Markdown can arrive from untrusted CLI/pipe/agent/file sources and was injected via `dangerouslySetInnerHTML` in the privileged renderer. Now sanitized with **DOMPurify** (forbids `javascript:` URIs, event handlers, `style`/form controls).

### markdown.load_file path abuse
Even with a valid token, the loader now enforces an extension allowlist (`.md`/`.markdown`/`.mdx`/`.txt`/`.text`/`.rst`), a 5MB cap, and an `isFile` check — so it can't be used to slurp secrets (`id_rsa`, `.env`, …) into the viewer.

### #9 — Webview hardening (Low/defense-in-depth)
Central `web-contents-created` handler: strips `preload`/`nodeIntegration` from attached webviews (enforces `contextIsolation`), routes `window.open` to the OS browser, and blocks the main window from navigating off its own UI (localhost/file://).

## Tests
- `tests/unit/pipe-server.test.ts`: token enforcement (reject without/with wrong token, accept correct token, public methods open, V1 still works) — 9 tests.
- `tests/unit/cdp-proxy.test.ts` (new): Host allowlist (loopback/IPv6/rebind/non-loopback) + Origin guard (absent/devtools allowed, web origins rejected) — 9 tests.
- Full suite: 106 passing. (`pty-manager.test.ts` fails only because the node-pty native module isn't built on this Linux dev box — pre-existing/environmental, fails identically on `master`.)

## Notes for reviewer
- CDP uses Host+Origin validation rather than a token because the external `chrome-devtools-mcp` connects via `--browserUrl=http://127.0.0.1:9222` (set in `claude-context.ts`) and can't send a custom token.
- The pipe token raises the bar from "any local code can drive the pipe" to "only processes running as this user (who can already read the user's files) can" — an honest, meaningful reduction for a single-user desktop app.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -19,6 +19,7 @@
         "@xterm/addon-web-links": "^0.11.0",
         "@xterm/addon-webgl": "^0.18.0",
         "@xterm/xterm": "^5.5.0",
+        "dompurify": "^3.4.11",
         "electron-updater": "^6.3.0",
         "marked": "^15.0.0",
         "node-pty": "^1.0.0",
@@ -2329,6 +2330,13 @@
         "@types/node": "*"
       }
     },
+    "node_modules/@types/trusted-types": {
+      "version": "2.0.7",
+      "resolved": "https://registry.npmjs.org/@types/trusted-types/-/trusted-types-2.0.7.tgz",
+      "integrity": "sha512-ScaPdn1dQczgbl0QFTeTOmVHFULt394XJgOQNoyVhZ6r2vLnMLJfBPd53SB52T/3G36VI1/g2MZaX0cwDuXsfw==",
+      "license": "MIT",
+      "optional": true
+    },
     "node_modules/@types/uuid": {
       "version": "10.0.0",
       "resolved": "https://registry.npmjs.org/@types/uuid/-/uuid-10.0.0.tgz",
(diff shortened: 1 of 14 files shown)


amirlehmam:
Merged and shipped in **v0.8.8** 🚀 Thanks @foschmitz — this is a substantial, well-documented hardening pass. The per-instance pipe token, CDP Host+Origin guards, DOMPurify markdown sanitization, `markdown.load_file` allowlist, and the central `web-contents-created` webview lockdown all landed. Your new `pipe-server` (token enforcement) and `cdp-proxy` (Host/Origin) test suites pass; `pty-manager` shows the same 5 pre-existing env failures you flagged (native built for Electron's ABI, not the Node test runner). The token file lives under the per-instance APPDATA dir and is injected into spawned shells as `WMUX_PIPE_TOKEN` for the CLI/hook.

PR #48, opened 2026-06-19
fix(terminal): restore tmux scroll and resize after browser pane open/close

### LINKED ISSUE #47: bug(terminal): tmux scroll and resize break when a browser pane is open
## Environment
- wmux: v0.8.7
- Shell: WSL / any shell running tmux

## Behavior

Two related bugs occur when a browser pane (split-tree webview) is open alongside a terminal running tmux:

**1. Mouse wheel scroll is stolen by the webview compositor**

When a `<webview>` element is present in the Electron window, Chromium routes un-prevented wheel events to the webview compositor instead of the terminal. The existing wheel handler correctly falls through on the alternate buffer (so xterm can handle it natively), but with a webview present that fall-through means the scroll event goes to the webview rather than tmux.

This is distinct from the general alternate-buffer regression tracked in #41 — it only occurs when a browser pane is simultaneously open.

**2. tmux does not redraw after browser pane close / terminal remount**

When the browser pane is closed, React remounts the terminal component. tmux doesn't re-send its mode-switching escapes (`\x1b[?1049h`) on SIGWINCH — it only does so on a fresh client attach. After remount, `xterm.buffer.active.type` reads `'normal'` even though tmux is still running, causing:
- The wheel handler to use the wrong scroll path
- `terminal.refresh()` in the ResizeObserver to fire unconditionally, painting stale buffer content before tmux's own SIGWINCH redraw arrives, leaving the terminal visually corrupted

Additionally, if the ResizeObserver fires before the `pty.create` IPC resolves (e.g. on fast remount), the resize event is silently dropped and tmux never receives SIGWINCH at all.

## Expected behavior

- Mouse wheel scrolls tmux content when a browser pane is open
- Terminal redraws correctly after browser pane close / component remount

## Notes

This is complementary to #41 / PR #42, which fix the general alternate-buffer scroll regression. This issue specifically requires handling the webview-compositor interception case.

schroldgames:
Fixes #47. Complementary to #41 / PR #42 — this addresses the specific case where a browser pane (webview) is open alongside a tmux terminal.

## Root causes

**Webview compositor stealing wheel events:** When a `<webview>` is present, Chromium routes un-prevented wheel events to it. The existing handler correctly falls through on the alternate buffer so xterm can handle it natively, but with a webview that fall-through sends the event to the webview instead of tmux.

**Unreliable buffer type after remount:** tmux doesn't re-send `\x1b[?1049h` on SIGWINCH — only on a fresh client attach. After React remounts the terminal (e.g. on browser pane close), `xterm.buffer.active.type` reads `'normal'` even though tmux is still running, causing the wrong scroll path and a premature `terminal.refresh()` that paints stale content before tmux's own SIGWINCH redraw arrives.

**Silent resize drop:** If ResizeObserver fires before `pty.create` IPC resolves, the resize is dropped and tmux never gets SIGWINCH.

## Changes (`src/renderer/hooks/useTerminal.ts`)

- **`surfaceMouseEnabled` map** (module-level, survives remounts): tracks SGR/button mouse enable (`\x1b[?100[0236]h`) and disable per surface. Used by the wheel handler to reliably identify tmux/vim sessions independent of `buffer.active.type`.

- **`hasWebviewCached` + MutationObserver**: caches webview presence to avoid a live `querySelector` on every wheel event. A single shared observer updates it when the DOM changes.

- **Wheel handler** now branches on webview presence and mouse-tracking state:
  | Condition | Action |
  |---|---|
  | No webview, normal buffer | Intercept → scrollback |
  | No webview, alt buffer | Fall through → xterm handles natively |
  | Webview + plain shell | Intercept → scrollback |
  | Webview + mouse-enabled (tmux/vim) | Intercept → write SGR scroll escapes to PTY |

- **`pendingResizeDims`**: captures ResizeObserver events that arrive before `pty.create` IPC resolves, flushed in `attachToPty`.

- **`doInitialResize`**: retries via RAF (up to 8 attempts) until `fitAddon.proposeDimensions()` returns non-null, ensuring tmux gets SIGWINCH and redraws into the new xterm instance after remount.

- **Conditional `terminal.refresh()` in ResizeObserver**: plain shells get it to mark rows dirty after a layout change; mouse-enabled apps skip it to avoid painting stale content before their own SIGWINCH redraw arrives.

## Relationship to PR #42

PR #42 fixes the general alternate-buffer scroll regression (#41) by taking ownership of wheel events in the alternate screen unconditionally. This PR fixes the orthogonal webview case — the two changes are complementary and can coexist without conflict.

### DIFF
--- a/src/renderer/hooks/useTerminal.ts
+++ b/src/renderer/hooks/useTerminal.ts
@@ -105,6 +105,24 @@ function buildXtermTheme(base: ThemeConfig, override?: UserColorScheme): ITheme 
 }
 
 const themeCache = new Map<string, ThemeConfig>();
+
+// Tracks whether mouse reporting is active for a given surface. Survives React
+// remounts so the wheel handler can distinguish tmux (mouse-enabled) from a
+// plain shell even when xterm's buffer.active.type is reset after remount.
+const surfaceMouseEnabled = new Map<string, boolean>();
+
+// Cache webview presence to avoid a live querySelector on every wheel event.
+// A single shared MutationObserver updates this whenever the DOM changes.
+let hasWebviewCached = false;
+let webviewObserverStarted = false;
+function ensureWebviewObserver(): void {
+  if (webviewObserverStarted) return;
+  webviewObserverStarted = true;
+  hasWebviewCached = !!document.querySelector('webview');
+  new MutationObserver(() => {
+    hasWebviewCached = !!document.querySelector('webview');
+  }).observe(document.body, { childList: true, subtree: true });
+}
 async function fetchTheme(name: string): Promise<ThemeConfig> {
   const cached = themeCache.get(name);
   if (cached) return cached;
@@ -198,6 +216,7 @@ export function useTerminal({ surfaceId, shell, cwd, visible = true, focused = t
 
     // Open terminal in the DOM
     terminal.open(terminalRef.current);
+    ensureWebviewObserver();
 
     // Always scroll wmux's buffer on wheel — never forward to the app.
     // Without this, when a TUI (Claude Code, vim, tmux…) enables mouse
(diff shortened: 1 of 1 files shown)


amirlehmam:
Merged and shipped in **v0.8.8** 🚀 Thanks @schroldgames!

This overlapped with #42 in `useTerminal.ts` (both rewrote the wheel handler), so I unified the two rather than taking one wholesale. The merged handler:
- **Always takes ownership on the capture phase** (your #47 webview fix + #42's #41 alt-buffer fix) — since we never fall through to xterm anymore, the `hasWebview` branching became redundant (a `preventDefault` on capture already stops the compositor stealing the event in *both* the scrollback and forward paths), so I dropped the `hasWebviewCached`/MutationObserver machinery as dead code.
- Keeps your reliable **`surfaceMouseEnabled`** signal (DECSET tracking that survives remounts) as the mouse-active discriminator, and #42's arrow-key fallback for non-mouse pagers (less/man).
- Keeps your `pendingResizeDims` + RAF-retried initial resize (extracted to module-level `scheduleInitialResize`/`scheduleDeferredRepaint` to stay within our complexity budget).

Net behavior matches your PR for every webview case, plus #42's reliability for the non-webview alt-buffer case. tsc + the full suite are green (the 5 `pty-manager` failures are pre-existing/environmental — node-pty's native is built for Electron's ABI, not the Node test runner).

PR #51, opened 2026-06-19
fix(pty): kill all PTYs on will-quit to prevent MSVC assertion crash

### LINKED ISSUE #50: bug(pty): MSVC assertion crash on close — remove_pty_baton(baton->id)
## Environment
- wmux: v0.8.7
- Windows (any)

## Behavior

When closing wmux, a Microsoft Visual C++ Runtime Library dialog appears:

```
Assertion failed!
Program: ...wmux.exe
File: ...
Expression: remove_pty_baton(baton->id)
```

The crash is intermittent — more likely when multiple terminal panes are open.

## Root cause

The `will-quit` handler stops ancillary services (pipe server, CDP proxy, port scanner, pollers) but never explicitly kills the PTY processes:

```typescript
app.on('will-quit', () => {
  pipeServer.stop();
  cdpProxy.stop();
  portScanner.stop();
  gitPoller.unwatchAll();
  prPoller.stopAll();
  // ← ptyManager.killAll() missing
});
```

When Electron exits without calling `ptyManager.killAll()`, node-pty's libuv async handles (called "batons") are still pending as the Node.js runtime tears down. The `remove_pty_baton` assertion fires because node-pty attempts to clean up those handles in a partially-destroyed state.

## Fix

Call `ptyManager.killAll()` as the **first** step in the `will-quit` handler, before any other teardown runs. This gives node-pty time to complete its cleanup before the process exits.

```typescript
app.on('will-quit', () => {
  ptyManager.killAll(); // ← add this
  pipeServer.stop();
  // ...
});
```

schroldgames:
Fixes #50.

## Problem

The `will-quit` handler tears down ancillary services but never calls `ptyManager.killAll()`. When Electron exits with PTY processes still live, node-pty's libuv async handles (batons) are pending as the runtime tears down, triggering:

```
Assertion failed! Expression: remove_pty_baton(baton->id)
```

## Fix

One line — call `ptyManager.killAll()` first in `will-quit`:

```typescript
app.on('will-quit', () => {
  ptyManager.killAll();
  pipeServer.stop();
  // ...
});
```

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -1016,6 +1016,10 @@ app.on('before-quit', () => {
 });
 
 app.on('will-quit', () => {
+  // Kill all PTYs before anything else tears down. Without this, node-pty's
+  // libuv async handles (batons) are still pending when the process exits,
+  // triggering the "Assertion failed: remove_pty_baton" MSVC runtime error.
+  ptyManager.killAll();
   pipeServer.stop();
   cdpProxy.stop();
   portScanner.stop();


amirlehmam:
Merged and shipped in **v0.8.8** 🚀 Thanks @schroldgames! One-line, high-value — `ptyManager.killAll()` is now the first thing `will-quit` does, ahead of the rest of the teardown, so the `remove_pty_baton` assertion (#50) is gone.

Issue #34, opened 2026-06-16
Redesign terminal toolbar with real SVG icons + Layout dropdown (fixes invisible new-tab dropdown)


kevmtt:
## Summary

Redesign the terminal pane toolbar (the `+ ▾ ⏐ ⎯ ×` controls) with **real inline SVG icons** and consolidate the two split buttons into a single **Layout** split-button with a dropdown. Also fixes a bug where the existing "new tab type" caret dropdown never appears.

All of this lives in one component: `src/renderer/components/SplitPane/SurfaceTabBar.tsx` (styled in `src/renderer/styles/splitpane.css`). `PaneWrapper.tsx` already wires up every handler needed (`onNew`, `onNewTyped`, `onNewProfile`, `onSplitRight`, `onSplitDown`, `onClosePane`), so this is a presentational redesign — no new plumbing.

## Bug fixed by this work

The "new tab type" caret (`▾`) currently does nothing when clicked, even with profiles configured. **Root cause:** `.surface-tab-bar` has `overflow: hidden` and is only 28px tall (`splitpane.css:52`), while the dropdown renders at `position: absolute; top: 30px` (`splitpane.css:151`) — i.e. below the bar — so it is clipped away and never visible. The click and state toggle work fine; the menu is just invisible.

**Fix:** render both dropdowns through a React **portal** to `document.body` with `position: fixed`, anchored to the trigger button's `getBoundingClientRect()`. This removes any dependency on parent `overflow` and works for both the New menu and the new Layout menu.

## Control cluster (right-aligned, ~10px gaps, rounded hover pills)

Compact split-button style: clicking the main icon performs the default action; the attached caret opens a dropdown.

| Control | Icon | Main click | Caret dropdown | Hover accent |
|---|---|---|---|---|
| **New** | plus inside a rounded square | open default terminal (`onNew`) | Terminal · Browser · Markdown — separator — quick-launch profiles | green |
| **Layout** | two side-by-side panes | split right (`onSplitRight`) | Split right `Ctrl+D` · Split down `Ctrl+Shift+D` | blue |
| **Close** | clean X | close pane (`onClosePane`) | — | red |

## Implementation notes

[...]

PR #43, opened 2026-06-18
feat(tabs): shell picker in caret dropdown + WSL home fix + dropdown overflow fixes

egertaia:
## Summary

Adds a **shell picker** to the new-tab caret dropdown, plus a WSL home-directory fix.

> **Stacked on #38.** This branch merges in the terminal-toolbar redesign (#38), so the diff includes its icon-based toolbar. The dropdown-clipping fix is provided by #38's portal-rendered menus — this PR layers the shell picker into that menu. **Merge #38 first**, then this PR shows only the shell-picker delta.

### This PR's own changes
- **Shell picker in the `▾` dropdown**: the caret next to `+` lists all detected shells (PowerShell 7, Windows PowerShell, WSL, Command Prompt) so you can open each in its own tab. Falls back to a generic Terminal entry if none are detected. Each shell spawns a PTY with its specific executable.
- **WSL starts in Linux home**: `wsl.exe` was being spawned with a Windows CWD, which made it mount into `/mnt/c/...`. Added `--cd ~` so WSL always starts in the Linux home directory.

### From #38 (merged in here)
- Real SVG icon toolbar (New / Layout split-buttons / Close).
- Portal-rendered dropdowns — fixes the menu being clipped by `.surface-tab-bar { overflow: hidden }` and by adjacent panes.

## Test plan

- [ ] Click `▾` caret — menu lists detected shells above Browser / Markdown
- [ ] Select WSL — new tab opens in Linux home (`~`), not `/mnt/c/...`
- [ ] Select PowerShell — new tab opens PowerShell
- [ ] Split into two panes, open the `▾` menu on the left pane — it renders on top of the right pane (not clipped)

## Notes
- Clipboard UTF-8 copy/paste fixes were split into a standalone PR (#44) — not part of this branch.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #38, opened 2026-06-17
feat(terminal): redesign terminal pane toolbar with real icons

kevmtt:
Presentational redesign of the terminal pane toolbar, plus a fix for the invisible new-tab dropdown.

Closes #34

## What & why

The toolbar used Unicode text glyphs (`+`, `▾`, split glyphs, `×`) that render inconsistently across fonts, and the "new tab type" caret dropdown was invisible.

- **Real icons** — replace glyphs with stroke-based inline SVG icons (`currentColor`, so hover accents come from CSS). New file `icons.tsx`.
- **Consolidated controls** — the two separate split buttons become a single **Layout** split-button (main icon = split right, caret = Split right / Split down).
- **Dropdown-clipping fix** — `.surface-tab-bar` is `height: 28px; overflow: hidden`, so the absolutely-positioned menu (`top: 30px`) was clipped and never painted. Menus now render via a **portal**, so both the New and Layout dropdowns are fully visible.

## Files

- `src/renderer/components/SplitPane/icons.tsx` — **new** SVG icon components (add, split, split-down, close, caret)
- `src/renderer/components/SplitPane/SurfaceTabBar.tsx` — icon split-buttons + portal-rendered menus
- `src/renderer/styles/splitpane.css` — control-cluster styling (rounded hover pills, accent colors), portal menu styles
- `docs/superpowers/specs/2026-06-16-terminal-toolbar-redesign-design.md` — design spec
- `.gitignore` — ignore the local `.superpowers/` tooling dir

## Scope

- No IPC / store / handler changes — `PaneWrapper` already passes `onNew`, `onNewTyped`, `onNewProfile`, `onSplitRight`, `onSplitDown`, `onClosePane`.
- Per-tab rendering, drag/drop, rename, and keyboard shortcuts unchanged.

## Test plan

- [x] `tsc` (renderer) clean
- [ ] New split-button: main icon opens a default terminal; caret shows Terminal / Browser / Markdown + profiles
- [ ] Layout split-button: split right (default) and Split down via caret
- [ ] Both dropdowns fully visible (no clipping) with quick-launch profiles configured
- [ ] Close button closes the pane

## Follow-up

Icons are hand-rolled SVGs for now; migrating to `lucide-react` is a sensible future step (noted on #34). The icons live in one file to make that swap trivial.

PR #52, opened 2026-06-19
fix(ui): caret dropdown clipped by overflow:hidden ancestors

schroldgames:
Fixes the bug documented in #34 (the dropdown is invisible due to clipping). This is a targeted fix without the broader toolbar redesign also proposed in that issue.

## Root cause

The new-tab type menu (▾ caret next to +) renders `position: absolute; top: 30px` inside `.surface-tab-bar`, but three ancestor elements have `overflow: hidden`:

- `.surface-tab-bar` (28px tall — the menu immediately clips below it)
- `.pane-wrapper__content`
- `.split-child`

`z-index` has no effect on `overflow: hidden` clipping. The click and state toggle work correctly — the menu is just never visible.

## Fix

Render the dropdown via `createPortal` at `document.body` with `position: fixed` + coordinates from `getBoundingClientRect()`. This escapes all parent clipping contexts entirely.

The outside-click handler is also tightened to exclude the caret button itself, preventing a mousedown-then-click double-toggle flicker when dismissing the menu by clicking the caret again.

## What this does not include

The SVG icon redesign and Layout split-button proposed in #34 are intentionally out of scope — those are a separate presentational change. This PR only fixes the clipping bug so the existing menu actually works.

### DIFF
--- a/src/renderer/components/SplitPane/SurfaceTabBar.tsx
+++ b/src/renderer/components/SplitPane/SurfaceTabBar.tsx
@@ -1,4 +1,5 @@
 import React, { useState, useEffect, useRef, useCallback } from 'react';
+import { createPortal } from 'react-dom';
 import { SurfaceRef, SurfaceId, PaneId, WorkspaceId, QuickLaunchProfile } from '../../../shared/types';
 import { useStore } from '../../store';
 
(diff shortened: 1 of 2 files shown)


kevmtt:
I fixed and reworked the design in this PR: https://github.com/amirlehmam/wmux/pull/38

amirlehmam:
Thanks @schroldgames — your diagnosis here is spot-on: three `overflow: hidden` ancestors clip the absolutely-positioned menu, and `z-index` can't escape a clipping context. 

Closing as **superseded**, because we landed the exact same fix via a different PR: **#38** (shipped in **v0.8.8**) renders the new-tab/Layout dropdowns through `createPortal` to `document.body` with `position: fixed` + `getBoundingClientRect()` coords — identical root-cause fix to yours — as part of the broader icon-toolbar redesign that closes #34. **#43** then layers a shell picker into those portal-rendered menus, and adds the same outside-click/right-edge-flip hardening you called out.

So the clipping bug is fixed in 0.8.8; your standalone targeted version would now conflict with the redesigned `SurfaceTabBar.tsx`. Really appreciate the clean writeup — and your #48 and #51 both shipped in this same release. 🙏

PR #61, opened 2026-06-25
[codex] Add live layout preview for surface dragging

AxelM-MSK:
﻿## Summary

Adds a live layout preview while dragging Surface tabs, so geometry-changing drops show the resulting workspace layout before the real split tree changes.

The preview is rendered as a lightweight overlay and keeps the real terminal, browser, markdown, and diff surfaces mounted underneath until drop. The implementation also covers source-pane collapse previews, center-drop previews only when layout changes, zoom cleanup on valid geometry-changing drops, and shared surface label resolution so preview tabs match real tab titles.

## Why

Before this change, dragging a Surface to an edge only highlighted the current drop zone. It did not show the full post-drop layout, which made source-pane collapse and multi-pane layouts harder to predict.

## Testing

- `npx.cmd vitest run tests/unit/surface-label.test.ts tests/unit/surface-drag-preview.test.ts tests/unit/split-preview.test.ts tests/unit/split-tree.test.ts`
- `npm.cmd run build:main`
- `npm.cmd run build:renderer`

I also ran `npm.cmd test`; it currently fails in the existing `tests/unit/pty-manager.test.ts` suite on this Windows checkout with 5 PTY-manager failures. The new renderer/split-preview tests pass, and this PR does not touch PTY manager code.

## Notes

Opening as draft for maintainer feedback on the interaction model and scope.

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -1,0 +1,17 @@
+# wmux
+
+wmux is a Windows terminal multiplexer organized around workspaces, split panes, and draggable surfaces. This glossary keeps product language precise when discussing layout and terminal interaction behavior.
+
+## Language
+
+**Surface**:
+A single terminal, browser, markdown, or diff instance shown as a tab inside a pane. Users drag a surface by grabbing its tab title.
+_Avoid_: Console window, terminal window, panel
+
+**Pane**:
+A rectangular region in a workspace split tree that contains one or more surfaces.
+_Avoid_: Console window, tab
+
+**Live Layout Preview**:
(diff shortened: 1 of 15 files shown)


amirlehmam:
Merged and shipped in **v0.12.0** 🎉 — thanks @AxelM-MSK! The live layout preview is a really nice touch. The implementation is clean: pure preview helpers that never mutate the real split tree, RAF-throttled rebuilds, and drops still go through the existing (now better-validated) store actions. The pre-existing `pty-manager` test failures you flagged are environmental (node-pty native spawn on this Windows checkout) and unrelated to this PR — confirmed they're the same 5 on master.

PR #70, opened 2026-07-05
Upgrade Electron 33 (EOL) → 43 + refresh build toolchain (clears 30 npm-audit advisories)

tawman:
## Why

**Electron 33.4.11 is end-of-life** (reached EOL 2026-04-29) and receives no further Chromium/V8 security patches. Electron only supports the latest three majors, so 33 is well outside the window. This PR moves the app to the current supported major, **Electron 43** (Chromium 150, Node 24, N-API 10), and refreshes the build toolchain in the same pass.

Beyond the EOL fix, the upgrade **clears every open `npm audit` advisory**: the tree went from **30 advisories (3 critical, almost all in the dev/build toolchain)** to **0**, full and `--omit=dev`.

The Electron jump is a Chromium **M130 → M150** and **V8 13 → 15** move — 20 majors of engine, rendering, and security work — plus Node 20 → 24 in the main process.

## Scope (deliberately narrow)

This is an **upgrade PR, not a refactor**. To keep it reviewable and low-risk:

- **No feature changes.** Only what the upgrade requires: the dependency bumps, two small CDP-compat fixes, one browser-panel navigation fix, dead-lint-directive removal, and one stale test.
- **Pre-existing issues are left as-is.** The `eslint:recommended` debt (`no-empty`, `no-control-regex`) that ESLint 10 surfaces is **not** cleaned up here — it predates this change and isn't upgrade-related. Likewise, only the one test that broke against a drifted contract is touched; **other pre-existing test/lint issues are out of scope**.
- Native modules: **node-pty needs no rebuild** — it ships ABI-stable N-API prebuilds, verified loading + spawning a PTY under Electron 43.

## What changed

**Runtime deps:** react/react-dom 19.2.7, zustand 5.0.14, ws 8.21.0, electron-updater 6.8.9, marked 18.0.5.

**Build toolchain:** vite 8.1.0 + @vitejs/plugin-react 6.0.3 + vitest 4.1.9 (coherent peer set), TypeScript 6.0.3, ESLint 10.6.0 (+ `@eslint/js`, now an explicit dep since ESLint 10 dropped the transitive hoist) with @typescript-eslint 8.62.1, electron-builder 26.15.3, concurrently 10, wait-on 9. `tsconfig` gains `ignoreDeprecations: "6.0"` for the TS 6 `moduleResolution: node` transition.

**Code:**
- `cdp-proxy` `/json/version` now derives Browser/UA/V8 from `process.versions` instead of a hard-coded `Chrome/133` — so it's correct under Chromium 150 and never goes stale on future Electron bumps.
- `cdp-bridge` full-page screenshot prefers `Page.getLayoutMetrics.cssContentSize` (current Chromium) with a fallback to the deprecated `contentSize`.
- **BrowserPane double-navigation fix:** `navigate()` set the `<webview>` `src` attribute *and* called `loadURL()` with the same URL, so every navigation fired two competing loads that aborted each other — logging a spurious `ERR_ABORTED loading <url>` from the main process on startup / dev-server auto-navigation. `src` is now the initial page only; all navigation goes through a single `loadURL`.
- Removed 9 dead `react-hooks/exhaustive-deps` disable directives (the plugin isn't registered, so they were no-ops that ESLint 10 now errors on).
- Fixed one test that used `PtyManager.create()`'s return as a bare id after it became an object.

## A few deliberate pins (why not "just latest everything")

Two packages are **intentionally held back**, because "latest" would break or over-reach:

- **uuid held at 11.1.1** — uuid **≥ 12 is ESM-only** (no `require` export). The main process is compiled to CommonJS, so `require('uuid')` would throw `ERR_REQUIRE_ESM`. v11 gives identical `v4()` behavior with a CJS entry, so there's no reason to force the ESM break.
- **@xterm/* held on the 5.x line** — xterm 6 is a coordinated major across the core + 8 addon packages; that's its own migration, not part of an Electron bump.

Everything else moves to the current release. Notable version moves:

| package | from | to | note |
|---|---|---|---|
| electron | 33.4.11 | 43.0.0 | EOL → supported (Chromium 150) |
| vite | 6 | 8.1.0 | plugin-react 6 needs vite ^8 |
| vitest | 2 | 4.1.9 | supports vite ^8 |
| @vitejs/plugin-react | 4 | 6.0.3 | |
| typescript | 5.5 | 6.0.3 | |
| eslint | 9 | 10.6.0 | + `@eslint/js` now explicit |
| @typescript-eslint/* | 8.58 | 8.62.1 | peers eslint ^10 + TS <6.1 |
| electron-builder | 25 | 26.15.3 | |
| marked | 15 | 18.0.5 | renderer-only, options compatible |
| react / react-dom | 19.0 | 19.2.7 | |
| **uuid** | 11 | **11.1.1 (held)** | v12+ is ESM-only; CJS main |
| **@xterm/\*** | 5.x | **5.x (held)** | xterm 6 = separate migration |

## Verification

- `npm ci` → **0 vulnerabilities** (full and `--omit=dev`).
- `build:main` (tsc) + `vite build` clean; ESLint runs (only the pre-existing recommended-rule debt noted above); **all tests pass**.
- Electron 43 boot probe: `process.versions` = electron 43.0.0 / chrome 150 / node 24; a `webviewTag` window loads; **node-pty loads and spawns a real PTY with no rebuild**.


amirlehmam:
Merged — thank you @tawman for another outstanding contribution! 🙏

The scoping on this one made it a pleasure to review: strictly the upgrade and what it forces, pre-existing debt explicitly left alone, and the two holds (uuid 11 for CJS, xterm 5.x as its own migration) called out with reasoning instead of silently pinned. The `process.versions`-derived `/json/version` is a nice touch — that hard-coded `Chrome/133` was a time bomb, and now future Electron bumps get CDP identification for free. Same for the BrowserPane double-navigation fix; those spurious `ERR_ABORTED` logs had been noise for a while.

Verified locally on top of master: `npm ci` clean (0 advisories), `build:main` + Vite 8 build green, all 158 tests passing, and node-pty prebuilds (incl. `conpty.dll`) loading fine under Electron 43 — the prebuild switch from #69 paid off immediately here, no rebuild needed.

This is now shipped as **[v0.16.0](https://github.com/amirlehmam/wmux/releases/tag/v0.16.0)**, with the release zip rebased onto the Electron 43.0.0 runtime (new Chromium 150 DLLs included).

PR #73, opened 2026-07-06
security: authenticate the V1 pipe surface and hook.event/agent.activity (fixes #72)

### LINKED ISSUE #72: Security: unauthenticated pipe surface lets any local process spoof agent status / notifications / diffs (agent.activity, hook.event, V1 all bypass the token)
## Summary

The control pipe (`\.\pipe\wmux`) authenticates *privileged* V2 methods with a per-instance token — good. But three unauthenticated paths let **any local process** mutate wmux's UI/agent state without ever presenting the token, and the pipe is created with **no restrictive ACL**, so the caller can be any process on the machine (any integrity level):

1. **All V1 text commands** — `PipeServer.handleV1` (`src/main/pipe-server.ts`) does **no** token check. Every V1 command is forwarded to all renderer windows as `METADATA_UPDATE` (`src/main/index.ts:377`), including `notify`, `report_pwd`, `report_pr`.
2. **`hook.event`** (in `PUBLIC_V2_METHODS`) — `src/main/index.ts:741` forwards arbitrary `params` to all windows as `HOOK_EVENT`, and for `tool: "Edit"|"Write"` fires `DIFF_UPDATE`. An anonymous caller can inject fake hook events and force diff-pane refreshes.
3. **`agent.activity`** (in `PUBLIC_V2_METHODS`) — `src/main/index.ts:762` calls `applyExternalActivity(surfaceId, …)` for an **arbitrary, caller-supplied `surfaceId`** with no ownership check, writing `lastTool` / `activeSkill` / `isDone` onto *any* surface's agent.

## Why this is a real problem

`pipe-server.ts` documents the allowlist principle as "read-only or telemetry-style … don't grant code execution / file access." But `agent.activity` is a **cross-surface state write** — process A can rewrite process B's agent status, e.g. mark a still-running agent `isDone: true`, or make an idle surface look busy. For a tool whose entire value proposition is giving a human a *trustworthy* view of what AI agents are doing, unauthenticated spoofing of agent state / notifications / diffs is an integrity issue:

- Fabricate "agent needs input" style notifications → phish the user into typing secrets into an attacker-chosen pane.
- Flip real agents' status (busy⇄done) so the operator loses track of what's actually running.
- Inject fake Edit/Write hook events and diff refreshes → misleading diff view.

Severity is local integrity/spoofing (not RCE), but it undermines the exact guarantee wmux exists to provide.

## Why the "telemetry must work tokenless" rationale doesn't hold

wmux already injects `WMUX_PIPE_TOKEN` into every spawned shell (`shared/instance.ts`), so the *legitimate* hook/agent-telemetry clients — Claude Code hooks, agents — run inside wmux shells and **have the token**. Keeping these methods in the tokenless allowlist only benefits processes that *aren't* wmux children, which is precisely the set that shouldn't be writing agent state.

## Proposed fix (contributable, code-level)

- **Move state-mutating methods out of `PUBLIC_V2_METHODS`** — at minimum `agent.activity` and `hook.event` should require the token. Keep only genuinely read-only `system.identify` / `system.capabilities` public.
- **Authenticate (or drop) V1 mutating commands** — require the token for `notify` / `report_pwd` / `report_pr`, or gate them behind the same allowlist logic.
- **Add an ownership check to `agent.activity`** so a caller can only update the surface it owns (e.g. match against `WMUX_SURFACE_ID`).
- **Restrict the named pipe with a current-user-only DACL** on creation, so non-user / low-integrity processes can't connect at all (defense in depth).

I can put up a PR that (a) trims the allowlist to the two read-only methods and (b) requires the token on the V1 mutating commands, since the real callers already carry it.

Rajveerx11:
Fixes #72.

Any local process could spoof agent status, notifications, and diff refreshes through three tokenless pipe paths: every V1 text command, `hook.event`, and `agent.activity`. Since wmux already injects `WMUX_PIPE_TOKEN` into every shell it spawns, all *legitimate* telemetry senders (shell integration, Claude Code hooks, the CLI, the opencode plugin via the CLI) already have the token — keeping these paths tokenless only benefited processes that were never supposed to write state.

## Changes

- **`src/main/pipe-server.ts`**
  - `PUBLIC_V2_METHODS` trimmed to the two genuinely read-only methods (`system.identify`, `system.capabilities`). `hook.event` and `agent.activity` now require the per-instance token like every other V2 method.
  - V1 lines accept an `auth <token> ` prefix (validated with the existing timing-safe `tokensMatch`). All V1 commands mutate UI state (`notify`, `report_pwd`, `report_pr`, shell state, …), so every command except the read-only `ping` now requires it. Unauthenticated lines get `unauthorized` back and are **not** forwarded to renderers.
- **`src/cli/wmux.ts`** — `sendV1` prepends the auth prefix (token already resolved from `WMUX_PIPE_TOKEN` / the APPDATA token file).
- **`src/shell-integration/wmux-powershell-integration.ps1`** — `Send-WmuxMessage` prepends the prefix, and the PR-polling background job receives the token via `-ArgumentList` and prefixes its `report_pr` line.
- **`tests/unit/pipe-server.test.ts`** — new coverage: V1 rejected without/with wrong token (and not emitted), `ping` still public, `hook.event`/`agent.activity` rejected tokenless and accepted with the token.

No client migration needed: `wmux-hook.ts` and the CLI's `sendV2` were already sending the token — it was just being ignored for these methods.

## Not included (deliberately)

- **Per-surface ownership check on `agent.activity`** — the token is per-instance, so the server can't verify which surface a caller owns without per-surface credentials. Requiring the token removes the *anonymous* spoofing this issue is about; per-surface tokens would be a separate, larger change.
- **Restrictive DACL on the pipe** — Node's `net.Server` doesn't expose a security descriptor for named pipes, so this needs a native module or a different pipe implementation. The token now gates every state write, which was the load-bearing gap.

## Testing

- `npm test` — 162/162 pass (13 in `pipe-server.test.ts`, 4 new).
- `npm run build:main` — clean.

### DIFF
--- a/src/cli/wmux.ts
+++ b/src/cli/wmux.ts
@@ -26,9 +26,11 @@ function readPipeToken(): string {
 const PIPE_TOKEN = readPipeToken();
 
 function sendV1(command: string): Promise<string> {
+  // V1 state updates authenticate with an "auth <token> " prefix (issue #72).
+  const line = PIPE_TOKEN ? `auth ${PIPE_TOKEN} ${command}` : command;
   return new Promise((resolve, reject) => {
     const client = net.connect({ path: PIPE_PATH }, () => {
-      client.write(command + '\n');
+      client.write(line + '\n');
     });
     let data = '';
     client.on('data', (chunk) => { data += chunk.toString(); });
--- a/src/main/pipe-server.ts
+++ b/src/main/pipe-server.ts
@@ -15,16 +15,18 @@ export interface V2Request {
   token?: string;
 }
 
-// V2 methods that are safe to call without authentication. These are read-only
-// or telemetry-style and don't grant code execution / file access. Everything
-// else (agent.spawn, browser.eval, markdown.load_file, pane/workspace mutation,
-// …) requires a valid per-instance token. Keeping an allowlist (rather than a
-// blocklist) means any new privileged method is locked down by default.
+// V2 methods that are safe to call without authentication. These are strictly
+// read-only and don't mutate any UI/agent state. Everything else — including
+// telemetry-style writes like hook.event and agent.activity, which can spoof
+// agent status / notifications / diff refreshes (issue #72) — requires a valid
+// per-instance token. The legitimate telemetry clients (Claude Code hooks,
+// agents, shell integration) all run inside wmux-spawned shells and carry
+// WMUX_PIPE_TOKEN, so nothing tokenless has a reason to write state. Keeping
+// an allowlist (rather than a blocklist) means any new privileged method is
+// locked down by default.
 const PUBLIC_V2_METHODS = new Set<string>([
   'system.identify',
   'system.capabilities',
-  'hook.event',
-  'agent.activity',
 ]);
 
 export interface V2Response {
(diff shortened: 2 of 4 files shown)


amirlehmam:
Merged and released in [v0.17.0](https://github.com/amirlehmam/wmux/releases/tag/v0.17.0) — thank you for a genuinely well-scoped security PR.

What I especially appreciated on review: the allowlist got *smaller* instead of the auth getting bolted on per-method, the PowerShell integration's background PR-poller job was not forgotten (easy to miss — it opens its own pipe connection), and the tests cover the exact spoofing scenario from the issue (`notify … agent needs your password` → `unauthorized`). All 162 tests green after merge.

The ownership check on `agent.activity` and the pipe DACL from your issue write-up are welcome as follow-ups.

PR #74, opened 2026-07-06
ci: wire SignPath Authenticode signing behind secrets, pin publisher, document updater kill switch (fixes #71)

### LINKED ISSUE #71: Security: auto-updater installs unsigned artifacts with no authenticity verification (quarantine window is the only defense)
## Summary

wmux ships **unsigned** release artifacts and the auto-updater performs **no cryptographic authenticity check** on what it downloads. The recently added 3‑day *quarantine window* (`src/main/updater.ts`) is currently the **only** defense against a malicious release — it buys detection time, but it does not verify *who* built the artifact. `updater.ts` itself flags this as an unshipped follow‑up:

```
// (Signature verification + Authenticode + build provenance are tracked as
//  follow-ups in the issue; they need offline keys / CI changes, not just code.)
```

## Where it stands in the code

- **`electron-builder.json`** — the `win` target has an icon but **no** `certificateFile` / `certificateSubjectName` / CSC config, so the produced `.exe`/NSIS/portable binaries are unsigned.
- **`.github/workflows/release.yml:58`** — the SignPath Authenticode signing steps are **entirely commented out** (`# ── Sign exe with SignPath (disabled — waiting for OSS approval) ──`).
- **`src/main/updater.ts`** — uses `electron-updater` with `autoDownload=false` + quarantine gating, but nothing calls into publisher-signature verification. On Windows, electron-updater can only verify a publisher signature when the build is actually Authenticode-signed; with unsigned builds that check is a no‑op.

## Why it matters

electron-updater's only integrity check on an unsigned Windows build is the SHA512 in `latest.yml` — but `latest.yml` is served from the **same** GitHub release as the installer, so it is self‑referential, not an authenticity guarantee. Anyone able to publish/replace a release (leaked maintainer token, compromised CI, or a future malicious-owner scenario) can ship a matching `latest.yml` + trojaned installer, and every client silently downloads it once the quarantine window elapses. That is effectively RCE-on-update for the whole install base, gated only by a 3‑day timer.

## Proposed fix (contributable, mostly config/CI)

1. Re-enable Authenticode signing in `release.yml` (uncomment/finish the SignPath OSS flow, or wire a self-managed cert via `CSC_LINK`/`CSC_KEY_PASSWORD`).
2. Add the signing config to `electron-builder.json` (`win.certificateSubjectName` / SignPath publisher) so `electron-updater` verifies the publisher on install.
3. Optionally pin an allowed publisher name and reject updates whose signature doesn't match, so the quarantine window becomes defense-in-depth rather than the sole control.
4. Document `WMUX_DISABLE_UPDATER=1` as the interim mitigation for security-sensitive users until signing lands.

Happy to open a PR wiring up the electron-builder signing config + the release.yml steps behind the existing SignPath secrets.

Rajveerx11:
Fixes #71.

Release artifacts are unsigned and the auto-updater has no authenticity check — the 3-day quarantine window is currently the only defense. This PR wires the signing pipeline so it activates the moment the SignPath secrets exist, with no further code changes needed.

## Changes

- **`.github/workflows/release.yml`** — the commented-out SignPath steps are re-enabled, with three fixes over the original block:
  - Each step is gated on `SIGNPATH_API_TOKEN` / `SIGNPATH_ORGANIZATION_ID` being configured (mapped to job env, checked in step `if:`), so releases keep working unsigned while SignPath OSS approval is pending and signing turns on automatically once the secrets land.
  - The steps moved **after** rcedit. The original placement signed the exe *before* rcedit rewrote the PE (icon + version metadata), which would have invalidated the Authenticode signature on every release.
  - After downloading the signed artifact, the workflow verifies it with `Get-AuthenticodeSignature` and fails the release if the signature is not `Valid` — no silently shipping a bad artifact.
- **`electron-builder.json`** — `win.publisherName: ["SignPath Foundation"]` (the cert subject SignPath OSS signs with), so electron-updater's publisher verification is pinned for signed exe/NSIS update flows. Please double-check the exact CN once the first signed build exists.
- **`README.md`** — new *Updates & security* section documenting the quarantine window, the current signing status, and `WMUX_DISABLE_UPDATER=1` / `WMUX_MIN_RELEASE_AGE_DAYS` as the interim mitigation for security-sensitive users (the code for both already exists in `updater.ts`).
- **`src/main/updater.ts`** — header comment updated to reflect what is now wired vs. what remains open.

## Honest scope note

electron-updater can only Authenticode-verify **exe/NSIS** update artifacts. wmux currently updates via a **zip**, which cannot carry an Authenticode signature — so even with signing live, the quarantine window remains the primary *client-side* control until the update artifact format changes (tracked as the natural follow-up). Signing still hardens the chain now: SmartScreen reputation, a verifiable artifact for users, and publisher pinning ready for the day the installer flow lands.

## Testing

- YAML validated (`js-yaml`), `electron-builder.json` parses, `npm run build:main` clean, `npm test` 158/158.
- Workflow logic (guarded steps) is inert without the secrets — identical behavior to today's unsigned releases.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -16,6 +16,13 @@ jobs:
   build-sign-release:
     runs-on: windows-latest
 
+    # SignPath Authenticode signing (issue #71). The signing steps run only when
+    # both secrets are configured, so releases keep working while SignPath OSS
+    # approval is pending — and turn on automatically once the secrets land.
+    env:
+      SIGNPATH_API_TOKEN: ${{ secrets.SIGNPATH_API_TOKEN }}
+      SIGNPATH_ORGANIZATION_ID: ${{ secrets.SIGNPATH_ORGANIZATION_ID }}
+
     steps:
       - uses: actions/checkout@v4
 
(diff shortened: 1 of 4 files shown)


amirlehmam:
Merged and released in [v0.17.0](https://github.com/amirlehmam/wmux/releases/tag/v0.17.0) — thanks, and good catch moving the signing steps after rcedit; that would have bitten us on the very first signed release.

Two things I had to adjust post-merge, for the record:
1. **`win.publisherName` → `win.signtoolOptions.publisherName`** — electron-builder 26.15.3 rejects the top-level key (`configuration.win should be one of these: null`), it moved into `signtoolOptions` in v26.
2. **Non-fatal submission failures** — the first live run hit `400 — Yearly quota for artifact size has been exceeded` from SignPath OSS, which failed the whole release. The submit step is now `continue-on-error` (ships unsigned with a warning, matching your "releases keep working while approval is pending" intent), while the signature *verification* of a downloaded signed artifact stays fatal, per your design.

So: pipeline is live, secrets are set, and the first release after SignPath quota frees up will be signed automatically. Your honest scope note about the zip artifact not being Authenticode-verifiable client-side is captured in the release notes as the follow-up.

PR #79, opened 2026-07-09
feat(sidebar): self-healing workspace status + set-status --workspace verb

tawman:
## Problem

The sidebar workspace status is driven by `shellState`, which is written only by the in-pane shell integration (`report_shell_state`: `running` on command start, `idle` when the prompt returns). A shell that emits `running` but is killed before returning to its prompt — e.g. an agent's interactive TUI reaped at teardown — never emits the matching `idle`. Once the workspace has only non-terminal surfaces left (markdown/browser), nothing can clear it, so the row stays stuck on **"Running"** indefinitely. An idle-but-open Claude session hits the same wall when the PTY observer doesn't catch the completion marker.

Separately, there's no way for a driver outside a pane (e.g. an orchestration coordinator) to set a *named* workspace's status — `set-status`/`set-progress`/`log` carry no surface target and are dropped in the renderer.

## Changes

1. **Auto-heal a stuck "Running".** When a shell PTY exits (`useTerminal` `onExit`), reset the owning workspace's `shellState` to `idle` if it was still `running` — a shell whose PTY has exited can't be the running command. Clears the badge the moment a pane/agent is torn down.
2. **`wmux set-status --workspace <id> --state <idle|running|interrupted> [--text "<label>"]`.** Sets a named workspace's status from anywhere, including outside a pane (keyed on `workspaceId`, not `surfaceId`). `--text` renders as `Done: <label>` (green). Wiring: CLI verb → `workspace.set_status` broadcast in main → a renderer branch that resolves by `workspaceId` and updates `{ shellState, notificationText }`. The pre-existing positional `set-status <key> <value>` form is unchanged.

## Testing

- `npm run build` clean; `npm test` green.
- Verified on a local build: a workspace stuck on "Running" flipped to "Idle" the instant its agents' PTYs were reaped, and `set-status --workspace … --state idle --text "…"` painted the row `Done: …` from a separate terminal.

### DIFF
--- a/src/cli/wmux.ts
+++ b/src/cli/wmux.ts
@@ -486,7 +486,25 @@ const COMMANDS: Record<string, (args: string[]) => Promise<void> | void> = {
   'clear-notifications': async (args) => print(await sendV2('notification.clear', { id: args[1] })),
 
   // Sidebar
-  'set-status': async (args) => print(await sendV2('sidebar.set_status', { key: args[1], value: args[2] })),
+  'set-status': async (args) => {
+    // `set-status --workspace <id> --state <idle|running|interrupted> [--text "<label>"]`
+    // sets a named workspace's sidebar status from anywhere (works outside a
+    // pane, unlike the surface-scoped shell integration). Without --workspace it
+    // falls back to the legacy positional `set-status <key> <value>` form.
+    const workspaceId = getFlag(args, '--workspace');
+    if (workspaceId) {
+      const state = getFlag(args, '--state');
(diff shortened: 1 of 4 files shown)


amirlehmam:
Merged — thanks Todd! The stuck **Running** badge has been a long-standing annoyance, and resetting `shellState` on PTY exit is exactly the right place to heal it: a shell whose PTY is gone can't be the running command. The `set-status --workspace` verb also fills a real gap for orchestration drivers coordinating from outside a pane.

Shipping this in **v0.20.0** shortly.

Issue #49, opened 2026-06-19
bug(terminal): terminal buffer clears on any split tree restructure (pane add/remove/drag)


schroldgames:
## Environment
- wmux: v0.8.7 (confirmed on fresh install)
- Reproducible on any shell (bash, zsh, PowerShell — does **not** require tmux)

## Reproduction steps

1. Open wmux with a single terminal pane. Type some commands so there is visible text in the buffer.
2. Click any link in the terminal (or use the + caret to open a Browser tab). A second pane appears.
3. The original terminal pane is now **completely blank**. All scrollback history is gone.

Also reproducible by:
- Dragging a terminal surface from one pane group to another
- Closing a pane (the surviving pane's buffer clears)
- Any operation that restructures the split tree depth

## Expected behavior

The terminal should retain its scrollback buffer and visible content across split tree changes.

## Root cause

`SplitContainer` is a recursive component that renders differently based on whether its node is a leaf or a branch:

**Single pane (leaf node):**
```
SplitContainer(root=leaf) → <div.split-child> → <PaneWrapper key="paneA">
```

**After adding a second pane (branch node):**
```
SplitContainer(root=split) → <div.split-container>
  → <div.split-child> → <SplitContainer(leaf)> → <div.split-child> → <PaneWrapper key="paneA">
```

`PaneWrapper` moves from 3 levels deep to 6 levels deep with a **different parent component**. React's reconciliation algorithm matches children by key **within the same parent** — since the parent changed, the `key="paneA"` prop does not prevent an unmount + remount. A fresh `xterm.Terminal` instance is created with an empty buffer. The PTY process keeps running (the keep-alive design works correctly), but xterm has no memory of what it previously displayed.

This affects **all** terminal types:
- **Plain shells (bash/zsh/PowerShell):** Buffer is gone. Terminal appears blank until the user presses Enter to get a new prompt. Scrollback history is permanently lost for that session.
- **TUIs (tmux, vim, etc.):** `pty.resize()` sends `SIGWINCH` after remount, which triggers a full screen redraw. Content comes back after ~100–200ms. Scrollback is still lost, but the active screen is restored.

## Why it wasn't obvious before

For tmux users, the SIGWINCH redraw makes the terminal appear to "survive" the remount. The issue is most visible with plain shells, and it only manifests when actually restructuring the split tree (most common operations — like switching tabs or workspaces — do not trigger a remount).

## Suggested fix

[...]

Issue #30, opened 2026-06-08
Fixes  修复


chenxionglz:
Fixes  修复
Terminal cursor now tracks typing under the Canvas renderer (https://github.com/amirlehmam/wmux/issues/23). In TUIs like Claude Code the block cursor could lag behind the text you typed. The Canvas renderer repainted the cursor on the blink timer's animation frame, which gets throttled in wmux's busy Electron renderer. The cursor is now repainted synchronously on every move (mirroring the WebGL renderer), so it stays glued to the caret while blinking still works.

<img width="116" height="52" alt="Image" src="https://github.com/user-attachments/assets/f5696f19-25d0-406b-91c7-cd7cfb6b46dc" />

I'm using Windows 10, and I often encounter spaces or overwritten text appearing in the input box when typing Chinese or English. Sending the text works fine, but this issue persists. I don't know what's causing the problem. I don't have this issue when using the built-in terminal.

amirlehmam:
你好 @chenxionglz，再次感谢你的耐心反馈！🙏 这次我们找到了真正的根本原因。

**为什么 v0.8.2 / v0.8.3 没修好**：你的截图（`确 认` 中间出现的空格、被覆盖的旧文字）说明这不是输入法的问题，也不只是光标的问题 —— 因为**发送出去的内容是正常的**，而且**输入英文也会出现**。问题出在 wmux 绘制终端用的 xterm.js **Canvas 渲染器**：它在上游已被**废弃（deprecated）**，存在没人修的绘制 bug —— 宽字符（中文）行重绘时会留下空白格子，或残留旧的文字。之前的修复只是给 Canvas 打补丁，治标不治本。

**v0.8.4 的修复**：可见的终端面板现在改用**WebGL 渲染器**（与 VS Code 终端相同的渲染引擎，持续维护，正确处理中文宽字符）。WebGL context 只在面板可见时占用、隐藏时释放，所以不会再触发当初迫使我们改用 Canvas 的 Chromium 上限问题。

请下载 **v0.8.4** 试一下：
https://github.com/amirlehmam/wmux/releases/tag/v0.8.4

（解压前记得右键 zip → 属性 → 解除锁定/Unblock）

用搜狗输入法在 Claude Code 里输入中文和英文测试一下，确认没问题我再关闭这个 issue。如果还有异常，麻烦再截图告诉我。

---

Hi @chenxionglz, thanks again for your patience! 🙏 This time we found the real root cause.

[...]

PR #84, opened 2026-07-11
Upgrade xterm.js 5.5.0 → 6.0.0

tawman:
Follow-up to the Electron 33 → 43 upgrade: now that the Chromium side is current, this brings the terminal core up to date too. xterm.js 6.0.0 (Dec 2025) is the first major since 5.5.0 (Apr 2024) — it's the version VS Code's terminal has been running as continuous betas for the last year.

## Dependency changes

| Package | Before | After |
|---|---|---|
| `@xterm/xterm` | 5.5.0 | **6.0.0** |
| `@xterm/addon-canvas` | 0.7.0 | **removed** (renderer deleted upstream in 6.0) |
| `@xterm/addon-fit` | 0.11.0 | 0.11.0 (already latest) |
| `@xterm/addon-image` | 0.8.0 | 0.9.0 |
| `@xterm/addon-search` | 0.15.0 | 0.16.0 |
| `@xterm/addon-serialize` | 0.13.0 | 0.14.0 |
| `@xterm/addon-unicode11` | 0.8.0 | 0.9.0 |
| `@xterm/addon-web-links` | 0.11.0 | 0.12.0 |
| `@xterm/addon-webgl` | 0.18.0 | 0.19.0 |

All addon bumps are the 6.0-paired stables (published the same day as core 6.0.0).

## What we get

- **Synchronized output (DEC mode 2026)** — apps can batch screen updates atomically. This is the one that matters most for wmux: redraw-heavy agent TUIs (Claude Code) in multiple panes flicker/tear noticeably less. No wmux code needed, it's automatic once the app requests it.
- **Reworked viewport/scrollbar** (ported from VS Code) — smoother scrolling, fixes for scrollbar teleporting, selection clearing on resize, wrapped-line find.
- **Faster search** — new `SearchLineCache`, which the FindBar benefits from on large scrollbacks.
- **Memory-leak + perf fixes** (CoreBrowserService leaks, cheaper marker clearing / listener disposal, batched sorted-list ops). These matter a lot given wmux's keep-alive tab model — every tab holds a live `Terminal` instance.
- **Built-in OSC 52** — our custom handler still takes precedence (custom handlers run first), so no behavior change; it's just a safety net now.

### Future option: `@xterm/addon-progress`

6.0 also ships a new progress addon (OSC 9;4 — the sequence Windows Terminal uses for taskbar progress). That's a natural fit for orchestration: a long-running agent in a pane could report real progress that wmux surfaces in the sidebar / tab, rather than the binary running/idle we show today. Not included here — flagging it as a follow-up this upgrade unlocks.

## Breaking change handled: the Canvas renderer is gone

6.0 deleted the canvas renderer, and `@xterm/addon-canvas` has no 6.x release (still peer-pinned to `^5.0.0`). wmux used it as the middle tier when the WebGL budget was exhausted or a context was lost, so:

- `terminal-renderer.ts`: the visible-pane chain is now **WebGL → DOM**. Over-budget, WebGL init failure, and `onContextLoss` all fall back to xterm's DOM renderer, which got real perf work in 6.0.
- Deleted `force-sync-cursor.ts` and its unit test. That patch existed solely to force the *canvas* cursor layer to repaint synchronously (it only repainted via the blink timer's rAF, which got throttled in our multi-pane + webview renderer — issues #23/#30). WebGL draws the cursor inline every frame and the DOM renderer has no separate blink layer, so with canvas gone the patch has nothing to patch.

Worth noting the canvas tier was already the buggy one — it mispainted wide CJK chars and stale rows under load (#23, #30), so dropping it is a net win rather than a loss.

## Other 6.0 removals — checked, not used

`windowsMode`, `fastScrollModifier`, and the moved `overviewRulerWidth` are all unused in wmux, so no changes needed. The alt→ctrl+arrow keybinding hack was also removed upstream; nothing here depended on it.

I also verified against the 6.0.0 source that the Korean/CJK IME patch in `useTerminal.ts` still applies — `CompositionHelper` keeps the same internals (`_compositionPosition`, `_isSendingComposition`, `_dataAlreadySent`, `_coreService`, `_textarea`) and still defers the textarea read via `setTimeout(0)`, so the race it fixes is still there and the override still lands.

## Testing

- `npm test` green (161 tests), `npm run build` clean, lint at parity with master.
- Ran a packaged build and exercised it live: typing/echo round-trips, scrollback (wheel + deep buffer reads), pane splits, closing a sibling pane to force a remount (the serialize-addon scrollback restore path, #49), FindBar search, and multiple concurrent Claude Code agents in panes.
- Also ran a 3-agent orchestration end-to-end on this build (spawn → run → reap → collapse) with no rendering or teardown regressions.

One cosmetic note for reviewers: the new viewport renders the scrollbar as a VS Code-style fade-out overlay (visible while scrolling) instead of an always-on native bar. That's upstream's design, not a regression.


amirlehmam:
Merged — thanks Todd, this is a fantastic follow-up to the Electron 43 upgrade. Synchronized output (mode 2026) is exactly what multi-pane Claude Code sessions needed, and dropping the canvas tier is indeed a net win given it was the renderer behind #23/#30. Appreciated the verification that the CJK IME patch still lands against 6.0 internals — that saved a real regression hunt. The @xterm/addon-progress idea for orchestration panes is a great follow-up, noted for the roadmap. Shipping in the next release.

amirlehmam:
Follow-up shipped: your @xterm/addon-progress suggestion is live in **v0.23.0** (https://github.com/amirlehmam/wmux/releases/tag/v0.23.0). OSC 9;4 now renders as a progress strip on surface tabs, an aggregated bar in the sidebar workspace rows (error state trumps siblings, determinate values averaged), and the Windows taskbar button via setProgressBar — exactly the orchestration use case you sketched. Thanks again for flagging it.

PR #85, opened 2026-07-11
feat(agent): --replace-tab on agent spawn — no more orphaned idle tabs in orchestration panes

tawman:
Fixes the orphaned-tab clutter that every orchestration run leaves behind in agent panes.

## The problem

`wmux layout grid --count N` creates N panes, each with a default terminal surface (an idle shell). `wmux agent spawn --pane <p>` then **appends** the agent's surface to that pane. So every agent pane ends up with **two tabs**: the untouched idle grid shell, plus the agent's Claude TUI.

That idle shell is pure noise — nothing ever runs in it, it holds a live PTY, and it survives `agent kill` (killing the agent reaps its process but leaves the dead surface as a tab). After a few waves the panes are cluttered with inert shells, and the tab bar makes it genuinely hard to tell which tab is the live agent.

## The fix

New `--replace-tab` flag on `wmux agent spawn`. When the target pane has **exactly one surface**, that surface is a **terminal**, and it is **not itself an agent surface**, the spawned agent surface *replaces* it (the idle shell's PTY is killed) instead of being appended. The pane ends with a single tab: the agent.

Any other pane state — multiple tabs, a non-terminal surface, an existing agent surface — falls back to the current append behavior, so this can't clobber a pane a user is actually working in. The flag is opt-in; without it, behavior is byte-for-byte what it is today.

The bundled `wmux-orchestrator` `spawn-agents.sh` now passes `--replace-tab`, so orchestration panes are single-tab out of the box.

## Changes

- `src/cli/wmux.ts` — `--replace-tab` flag on `agent spawn`
- `src/shared/types.ts` — plumb the option through the spawn payload
- `src/main/index.ts` — pass it to the renderer bridge
- `src/renderer/App.tsx` — apply replace-vs-append at spawn time
- `src/renderer/store/split-utils.ts` — `replaceSurfaceInLeaf` helper (immutable, consistent with the rest of the split-tree mutations)
- `tests/unit/split-tree.test.ts` — coverage for the replace path and for each guard that falls back to append
- `resources/wmux-orchestrator/` — `spawn-agents.sh` passes the flag; skill doc updated

## Testing

- `npm test` green (172 tests, including the new split-tree cases covering the replace path and every fallback guard).
- Verified live on a packaged build with a 3-agent orchestration: from a single-pane start, `spawn-agents.sh` produced a clean 2×2 with **every agent pane at 1 tab** (previously 2 — grid shell + agent TUI), coordinator pane untouched at 1 tab, zero orphans. Full cycle exercised: spawn → agents run and report → `agent kill` (all agents `exited`, PIDs confirmed dead) → `close-pane` → collapse back to the single pane.
- Confirmed the guards: panes with multiple tabs / non-terminal surfaces / existing agent surfaces still append as before.

## Note for anyone on an older build

`agentSpawn` parses flags as `--flag value` pairs, so passing `--replace-tab` (a valueless flag) to a CLI that doesn't know it will shift the pairing and silently swallow the next flag. Only pass it to a build that has this change — worth knowing if the orchestrator plugin is ever updated independently of the app.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -377,7 +377,7 @@ wmux browser fill @eN <value> | get-text | screenshot | eval <js>
 wmux browser back | forward | reload
 
 # Agents
-wmux agent spawn [--cmd C] [--label L] [--cwd D] [--pane P]
+wmux agent spawn [--cmd C] [--label L] [--cwd D] [--pane P] [--replace-tab]
 wmux agent spawn-batch --json '[...]' [--strategy distribute|stack|split]
 wmux agent status <id> | list | kill <id>
 
--- a/resources/wmux-orchestrator/scripts/spawn-agents.sh
+++ b/resources/wmux-orchestrator/scripts/spawn-agents.sh
@@ -84,11 +84,14 @@ while IFS= read -r agent; do
 
   # launch-agent.js uses execFileSync with '--' separator to pass the prompt
   # as a positional arg — full interactive TUI, user can watch and intervene.
+  # --replace-tab: the agent surface takes over the grid pane's default idle
+  # terminal tab instead of being appended next to it (single-tab agent panes).
   SPAWN_RESULT=$(wmux agent spawn \
     --cmd "node \"$LAUNCHER\" \"$PROMPT_FILE\"" \
     --label "$AGENT_LABEL" \
     --cwd "$CWD" \
-    --pane "$PANE_ID" 2>&1)
+    --pane "$PANE_ID" \
+    --replace-tab 2>&1)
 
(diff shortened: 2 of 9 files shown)


amirlehmam:
Merged — thanks! The orphaned idle-shell tabs were a real eyesore after every orchestration run, and the guard conditions (single surface, terminal type, not already an agent) make this safe to enable by default in spawn-agents.sh. Also appreciated the heads-up about the --flag value pairing hazard on older builds; since the orchestrator plugin ships bundled with the app they stay in lockstep, but it's worth knowing. Going out in the next release.

PR #93, opened 2026-07-12
fix(orchestration): validate state.json before it reaches the renderer; add error boundary

### LINKED ISSUE #92: Malformed wmux-orch-*/state.json blanks the entire window (uncaught TypeError unmounts the React tree)
A single malformed `state.json` in the temp dir takes the whole wmux window black — permanently, on every launch, until the file is removed. No crash dialog, process stays alive, only a restart "fixes" it (and then it happens again).

Hit this three times in a row on 0.23.0 running an orchestration.

## Repro

Drop this in `%TEMP%\wmux-orch-repro\state.json` and start wmux:

```json
{
  "orchestrationId": "repro",
  "task": "malformed run",
  "status": "running",
  "waves": [{ "index": 0, "status": "running", "agents": [] }]
}
```

Note `orchestrationId` where the schema wants `id` — a plausible typo, and exactly what an agent hand-rolling the file produced for me. Window goes black within a second.

## Root cause

`orchestration-watcher.ts` reads state.json with a **cast, not a check**:

```ts
const parsed = JSON.parse(raw) as OrchestrationState;   // no validation
parsed._orchDir = orchDir;
return parsed;                                          // broadcast to renderer every 1s
```

state.json is written by a *different* process (the wmux-orchestrator plugin, or an agent), so it is untrusted input. The unvalidated object reaches `OrchestrationPanel`:

```tsx
{orch.id.replace(/^orch-/, "")}   // TypeError: Cannot read properties of undefined (reading "replace")
```

Two things turn that into a dead app:

1. **No error boundary anywhere in the renderer.** React unmounts the entire tree on an uncaught render throw, leaving the window painted in its `#1a1a1a` background — indistinguishable from a hang.
2. **The watcher prefers `status: "running"`.** A stale/malformed running file therefore wins selection on every poll tick, so it re-crashes on every launch.

The console shows exactly one line, then nothing — because after the unmount there is nothing left to re-render and throw again:

```
INFO:CONSOLE "Uncaught TypeError: Cannot read properties of undefined (reading 'replace')"
```

## Impact

Any process that can write `%TEMP%\wmux-orch-*\state.json` can hard-crash wmux. Worth noting `scripts/orchestration-state.sh` already carries trust-hardening comments (issue #2 / F-9) about not blindly following a `running` state.json — the watcher never got the same treatment.

## Fix

PR incoming: validate `id`/`status`/`waves` in the watcher before broadcasting, add an `ErrorBoundary` (root + orchestration panel) so no single data-driven panel can blank the window again, and guard the panel itself. Includes a regression test for the malformed shapes.

tawman:
Fixes #92 — a malformed `wmux-orch-*/state.json` blanks the entire window.

## The bug

`orchestration-watcher.ts` reads state.json with a cast, not a check:

```ts
const parsed = JSON.parse(raw) as OrchestrationState;   // no validation
return parsed;                                          // → broadcast to the renderer every 1s
```

state.json is written by a separate process (the plugin, or an agent hand-rolling it), so it is untrusted input. A run written with `"orchestrationId"` instead of `"id"` reaches `OrchestrationPanel`, which does `orch.id.replace(/^orch-/, "")` and throws. With **no error boundary anywhere in the renderer**, React unmounts the whole tree and the window is left painted in its `#1a1a1a` background — looks like a hang, cleared only by a restart. And since the watcher *prefers* `status: "running"`, one stale file reproduces it on every launch.

Repro and full analysis in #92.

## The fix

- **`orchestration-watcher`** — shape-check `id` / `status` / `waves` before broadcasting; ignore anything that fails, warning once per directory rather than once per 1s poll tick.
- **`ErrorBoundary`** (new) — wraps the root as a last-resort net, and the orchestration panel specifically (silent, so a bad run just hides the panel). This kills the bug *class*: no single data-driven panel can blank the window again. Reusable for the other panels that render data wmux does not own (session snapshots, agent metadata).
- **`OrchestrationPanel`** — bail out on a run with no `id`/`waves` instead of throwing.

## Verification

- New `tests/unit/orchestration-watcher.test.ts` pins the exact `orchestrationId`-instead-of-`id` shape, plus missing/empty/non-string `id`, missing `status`, non-array `waves`, and malformed waves.
- `npm test` → 197 passing (26 files). `tsc` clean. No new lint problems (35 on master, 35 on this branch).
- Verified live in a packaged build: planting the malformed state.json above with `status: "running"` no longer produces a console error and the window stays up; a well-formed run still renders in the sidebar as before.

### DIFF
--- a/src/main/orchestration-watcher.ts
+++ b/src/main/orchestration-watcher.ts
@@ -44,16 +44,54 @@ function listOrchDirs(): string[] {
   }
 }
 
+/**
+ * Shape-check a parsed state.json before it is allowed near the renderer.
+ * state.json is written by a separate process (the Claude Code plugin, or an
+ * agent hand-rolling the file), so it is untrusted input. A run missing `id`
+ * or `waves` used to be broadcast anyway and then throw inside the sidebar's
+ * render — and an uncaught throw in render unmounts the entire React tree,
+ * leaving a black window that only a restart clears. Anything that does not
+ * satisfy the contract is ignored rather than shipped to the UI.
+ */
+export function isValidState(value: unknown): value is OrchestrationState {
+  if (!value || typeof value !== 'object') return false;
+  const s = value as Record<string, unknown>;
+  if (typeof s.id !== 'string' || s.id === '') return false;
+  if (typeof s.status !== 'string') return false;
+  if (!Array.isArray(s.waves)) return false;
+  return s.waves.every((w) => {
+    if (!w || typeof w !== 'object') return false;
+    const wave = w as Record<string, unknown>;
+    return typeof wave.index === 'number' && Array.isArray(wave.agents);
+  });
+}
+
+// Dirs we've already rejected, so a bad file doesn't log once per poll tick.
+const warned = new Set<string>();
+
 function readState(orchDir: string): OrchestrationState | null {
   const stateFile = path.join(orchDir, 'state.json');
+  let parsed: unknown;
   try {
-    const raw = fs.readFileSync(stateFile, 'utf-8');
-    const parsed = JSON.parse(raw) as OrchestrationState;
-    parsed._orchDir = orchDir;
-    return parsed;
+    parsed = JSON.parse(fs.readFileSync(stateFile, 'utf-8'));
   } catch {
+    // Missing, unreadable, or mid-write (partial JSON) — try again next tick.
     return null;
   }
+
+  if (!isValidState(parsed)) {
+    if (!warned.has(orchDir)) {
+      warned.add(orchDir);
+      console.warn(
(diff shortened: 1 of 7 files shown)


amirlehmam:
Merged — thanks for another surgical fix. The analysis in #92 was spot on: the watcher trusting a cross-process file and the renderer having no error boundary were two separate holes, and this closes both. The `silent` boundary around the orchestration panel + the root-level net is exactly the right layering, and pinning the `orchestrationId`-instead-of-`id` shape in the regression test means this exact failure can't come back. Also appreciated the drive-by fix on the epoch-elapsed rendering ("495520:02:09"). Ships in the next release.

Issue #60, opened 2026-06-25
WSL Base Support for notification framework


stlindenau:
Currently the notification framework (in the sidebar) is not really working in wmux. 
While it would be a bigger effort to get a seamless integration the following 2 things would already improve the situation a lot

1. Ensure that WMUX variables are pushed through - the following environment variable worked on my side when set manually: 
~~~
WSLENV="WMUX/u:WMUX_SURFACE_ID/u:WMUX_CLI/up:WMUX_PIPE/u"
~~~

2. Translate WSL workspace path on session restore - currently an error is showing up if a WSL path like /home/<userid>/.../ is put into the variable windows.workspaces.cwd in session.json - the following error is the result:
~~~
[failed to create PTY: Error: Error invoking remote method 'pty:create': Error: Failed to create terminal: Cannot create process, error code: 267]
~~~

These 2 fixes/enhancements would make working with WSL a lot easier. 

Thanks in advance,
Stefan Lindenau

amirlehmam:
Fixed in **v0.12.0** 🎉 — thanks for the precise report and the `WSLENV` hint, Stefan, it made this straightforward.

**1. WMUX vars now propagate into WSL.** wmux sets `WSLENV` automatically for WSL shells:
```
WMUX/u:WMUX_SURFACE_ID/u:WMUX_CLI/up:WMUX_PIPE/u:WMUX_PIPE_TOKEN/u:WMUX_INTEGRATION/u
```
(`/up` translates the `WMUX_CLI` Windows path to a `/mnt/c/...` mount; the rest pass through unchanged.) Any pre-existing `WSLENV` you set is preserved and appended to. So the sidebar/notification framework and the `wmux` CLI now work from inside WSL without manual setup.

**2. WSL cwd on session restore (error 267).** A restored POSIX path like `/home/<user>/...` was being handed to a Win32 process as its working dir, which fails with `ERROR_DIRECTORY (267)`. wmux now detects a POSIX cwd and instead opens it *inside* the distro via `wsl --cd <path>`, while the Win32 process starts in a valid Windows dir. So restoring a session with a WSL workspace folder no longer errors.

Grab the v0.12.0 zip from the releases page (right-click → Unblock after download). Let me know if anything's still off on your setup!

stlindenau:
@amirlehmam  thanks for the quick reaction. The environment variables are working neatly.


[...]

PR #94, opened 2026-07-12
fix(pty): fall back to a real directory when the cwd no longer exists (spawn error 267)

tawman:
A pane dies with an opaque native error when its working directory does not exist:

```
Failed to create terminal: Cannot create process, error code: 267
```

267 is `ERROR_DIRECTORY` — CreateProcess was handed a working dir that is not a real directory. node-pty surfaces it verbatim, with a native stack trace and no indication of which path was at fault.

## Cause

`pty-manager` already guarded **one** route to 267 — a POSIX/WSL cwd restored from session.json (issue #60):

```ts
const spawnCwd =
  options.cwd && isPosixPath(options.cwd) ? (process.env.USERPROFILE || "C:\\") : options.cwd;
```

But a plain Win32 cwd was passed straight through, unchecked. It arrives from session state and CLI args, so it cannot be trusted to still exist at spawn time. Hit this repeatedly during a long orchestration run: agents are spawned into git worktrees, and a worktree that was removed after its wave — or whose `git worktree add` had not finished yet — leaves the spawn pointing at a directory that is not there.

The existing `catch` compounds it: on failure it retries with `useConptyDll: false`, a fallback meant for conpty **load** failures. With a bad cwd that retry hits the same missing directory, throws 267 again, and this time nothing catches it.

## Fix

Resolve the cwd before spawning (`resolveSpawnCwd`): keep it when it is a directory that exists, otherwise fall back to `USERPROFILE` and warn with the offending path. Folds the existing issue-#60 POSIX case into the same helper, and also rejects a cwd that exists but is a file.

A pane that would have died now opens in the home directory with a clear reason in the log.

## Verification

- 6 new cases in `tests/unit/pty-manager.test.ts`: cwd that exists, deleted cwd (the worktree case), never-existed cwd, cwd that is a file, POSIX/WSL cwd, and `undefined` passthrough.
- `npm test` → 195 passing (25 files). `tsc` clean. No new lint problems (4 pre-existing warnings in `pty-manager.ts`, unchanged).

### DIFF
--- a/src/main/pty-manager.ts
+++ b/src/main/pty-manager.ts
@@ -129,6 +129,30 @@ function isPosixPath(p: string): boolean {
   return p.startsWith('/') && !p.startsWith('//');
 }
 
+// Resolve the working dir handed to pty.spawn, guaranteeing it is a directory
+// that exists — otherwise CreateProcess fails with error 267 (ERROR_DIRECTORY)
+// and the pane dies with an opaque "Cannot create process, error code: 267".
+// Returns undefined (node-pty's own default) when there is nothing usable.
+export function resolveSpawnCwd(cwd: string | undefined): string | undefined {
+  if (!cwd) return undefined;
+
+  const fallback = process.env.USERPROFILE || 'C:\\';
+
+  // POSIX/WSL cwd: not a valid Win32 working dir at all (issue #60).
+  if (isPosixPath(cwd)) return fallback;
+
+  // Win32 cwd that no longer exists (deleted git worktree) or does not exist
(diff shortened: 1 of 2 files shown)


amirlehmam:
Merged — thanks! Folding the issue-#60 POSIX case and the new deleted-worktree case into one `resolveSpawnCwd` helper is cleaner than stacking guards, and rejecting a cwd that exists-but-is-a-file is a nice catch. You're right that the `useConptyDll: false` retry was masking the real failure mode on the second throw. A pane opening in $USERPROFILE with a warning beats a dead pane with "error code: 267" every time. Ships in the next release.

Issue #68, opened 2026-07-04
[Bug] v0.15.0 release missing latest.yml — auto-updater throws 404 on every launch


Dr-ilies:
# [Bug] v0.15.0 release missing `latest.yml` — auto-updater throws 404 on every launch

## Summary

Starting with v0.15.0, wmux's bundled `electron-updater` throws a noisy 404 on every launch because the GitHub release for v0.15.0 is missing the `latest.yml` metadata file that electron-updater requires.

This is purely cosmetic (the app still works), but the error is printed on startup and every 6 hours thereafter (`RECHECK_INTERVAL_MS = 6 * 60 * 60 * 1000` in `src/main/updater.ts`).

## Environment

- **wmux version:** v0.15.0
- **OS:** Windows 10/11
- **Install method:** Release zip (`wmux-0.15.0-win-x64.zip`)
- **Install location:** `%LOCALAPPDATA%\wmux\`

## Steps to reproduce

1. Download `wmux-0.15.0-win-x64.zip` from https://github.com/amirlehmam/wmux/releases/tag/v0.15.0
2. Extract anywhere
3. Run `wmux.exe`

## Observed output

```
[wmux] Installed wmux OpenCode plugin to C:\Users\ilies\.config\opencode\plugin\wmux.js
Checking for update
wmux pipe server listening on \\.\pipe\wmux
Error: Error: Cannot find latest.yml in the latest release artifacts
(https://github.com/amirlehmam/wmux/releases/download/v0.15.0/latest.yml): HttpError: 404
  "method: GET url: https://github.com/amirlehmam/wmux/releases/download/v0.15.0/latest.yml
   Please double check that your authentication token is correct..."
[updater] Auto-updater error: Error: Cannot find latest.yml in the latest release artifacts...
  code: 'ERR_UPDATER_CHANNEL_FILE_NOT_FOUND'
```

## Root cause

Looking at the v0.15.0 release: https://github.com/amirlehmam/wmux/releases/tag/v0.15.0

The release ships:
- `wmux-0.15.0-win-x64.zip` (the binary)
- `Source code (zip)` / `Source code (tar.gz)` (auto-generated by GitHub)

[...]

Issue #32, opened 2026-06-14
Feedback from a Windows user: notification sounds, installer/Store distribution, CLI new-instance bug, and per-pane quick-launch profiles


kevmtt:
First off — I really love that this tool exists and lets Windows users catch up to macOS developers. I've been testing wmux for my work and wanted to share constructive feedback on a few things I struggled with, plus one feature suggestion. Code references included in case they help.

---

### 1. Notification sounds never play

The Sound setting (Settings → Notifications, `Default`/`None`) looks functional but is a no-op:
- `notificationPrefs.sound` is defined (`settings-slice.ts:246-260`) and shown in the UI (`NotificationSettings.tsx:51-65`) but **never read anywhere**.
- There's no audio playback code at all (no `new Audio`/`.play()`), and `resources/sounds/` is empty (only `.gitkeep`) despite being bundled into every release.
- Only the Electron toast + taskbar flash + pane ring fire (`notification-manager.ts:4-14`).

**Suggestion:** ship a default sound asset and play it in the `NOTIFICATION_FIRE` path when `sound !== 'none'`, or remove the setting until implemented so it's not misleading.

---

### 2. Proper installer / Microsoft Store distribution + PATH

The portable zip works but limits discoverability and makes PATH setup manual. `electron-builder.json` already has NSIS config (currently unused) and electron-builder supports MSIX/`appx`. Rough effort, lowest-friction first:

| Option | Est. effort | Cost / prereqs | Notes & risks |
|---|---|---|---|
| **Winget** (`microsoft/winget-pkgs`) | **~0.5–1 day** | Free. No code-signing required to publish. | Lowest-friction win — a YAML manifest pointing at the existing release zip/exe + SHA256, fully automatable per release via the `winget-releaser` GitHub Action. Caveat: unsigned binaries still trip SmartScreen on *first run*, so this improves install UX, not trust. |
| **NSIS installer** (already configured) | **~1–2 days** | Code-signing cert is the real cost (below). | Config mostly exists; work is building off the OneDrive/spaces path (the documented `winCodeSign` blocker), wiring signing into the release pipeline, and testing shortcut/PATH registration. **Unsigned, NSIS trips SmartScreen *more* than the zip** — only pays off once signing is in place. |
[...]

PR #97, opened 2026-07-14
fix(updater): ship signed NSIS installer so Windows auto-update installs (#96)

### LINKED ISSUE #96: Auto-updater downloads but never installs on Windows — app stays on old version
**I am facing this issue myself** on Windows 11 — my installed wmux never updates. When a new release comes out I get the update notification, click it, the app restarts, and it's still on the old version. The first update never lands and it just keeps prompting me.

After digging into it, here's what's actually going on:

### The problem
On Windows the app can *detect and download* an update fine, but it can never *install* it. So it quits, relaunches the same old version, and the update badge shows up again — forever.

### Why it happens
- The Windows release only ships `latest.yml` + a `.zip` (`wmux-x.y.z-win-x64.zip`). There's no NSIS `.exe` installer or `.blockmap`.
- electron-updater on Windows installs by running an NSIS installer. A zip is a macOS update mechanism, so `quitAndInstall()` has nothing to run and effectively no-ops. The download succeeds and the checksum verifies, which is why it *looks* like it's working — but nothing ever gets applied.
- Two smaller things pile on: the update badge in the titlebar just opens the GitHub release page (it doesn't trigger the real installer), and the 3-day quarantine window in `updater.ts` often blocks fresh releases from even downloading since releases ship more than once a day.

### How I'd like to fix it
- Switch the Windows build target to `nsis` (the `nsis` block already exists in `electron-builder.json`, it's just not referenced), build without `--dir`, and let electron-builder generate `latest.yml` + `.blockmap` instead of hand-writing the yaml in the release workflow.
- Wire the in-app update badge to the real install flow (or remove the duplicate path) so clicking it actually does something.
- Make the quarantine window a bit more sensible so updates aren't held indefinitely.

Happy to work on this if you'll assign it to me — I've already traced it end to end.

Rajveerx11:
Closes #96.

## The problem
On Windows the app detects and downloads updates, but can never install them. The release ships a `.zip` (`wmux-<ver>-win-x64.zip`) as the update artifact. electron-updater on Windows uses `NsisUpdater`, which installs by running an NSIS `.exe` — it has no path to apply a zip. So the download and checksum both succeed (it *looks* like it's working), but `quitAndInstall()` no-ops and the app relaunches on the same version. The updater's own comment already flags the zip as the acknowledged follow-up (#71).

## The fix
Ship a real NSIS installer whose hashes match `latest.yml`:

- **`electron-builder.json`** — switch `win.target` from `dir` to `nsis` (the `nsis` block was already configured) and add a space-free `artifactName` (`wmux-${version}-setup.exe`).
- **`release.yml`** — build the installer (drop `--dir`), retarget the SignPath signing steps at the installer, remove the manual `Compress-Archive` zip, and generate `latest.yml` **last, from the signed installer**, so its `sha512` matches the exact file that ships. The redundant `rcedit` step is removed — electron-builder already embeds `win.icon`, the version, and `FileDescription` (taskbar-pinning name) into the packaged exe.
- **Regression test** — asserts the Windows target is `nsis`, not `dir`/zip, so this can't silently regress.

## Respecting the signing sequencing (#32)
Per #32, an NSIS installer isn't worth shipping until code-signing is live, since (a) unsigned NSIS trips SmartScreen harder than a zip, and (b) with `win.publisherName` pinned to `SignPath Foundation`, electron-updater now Authenticode-**verifies** the downloaded installer — an *unsigned* NSIS installer would fail that check and the update would throw.

So this PR is designed to **degrade gracefully** and is gated behind active signing:

- `latest.yml` is only published when the SignPath submission succeeds (the verify step is already fatal). Unsigned releases still ship the installer for **manual download**, but don't advertise an auto-update — a no-op, not a broken update. (`updater.ts` already treats a missing `latest.yml` as "update check skipped", #68.)
- Once the SignPath OSS quota is restored (#71), signing — and therefore verifiable auto-update — turns on automatically with **no further changes**.

Net: merge-safe now (auto-update simply stays dormant while unsigned), and it starts working the moment signing is live.

## Notes for review
- **SignPath config:** the submit still uses `artifactConfigurationSlug=initial`, which was set up for the old unpacked `wmux.exe`. Submitting the installer `.exe` may need that artifact configuration updated on the SignPath side. It's non-fatal (`continue-on-error`), but worth checking when quota returns.
- **Build env:** NSIS is built in CI (`windows-latest`, clean path) — the `OneDrive - Pulsa` spaced-path winCodeSign issue from `CLAUDE.md` doesn't apply on the runner.
- No `.blockmap` is emitted, so updates are full downloads (verified by the top-level `sha512`) rather than differential — correct, just not delta-optimized.

## Testing
- `npm test` — full suite green (211 passed), including the new target guard.
- Verified every workflow reference resolves to `release/wmux-<ver>-setup.exe` (signing submit/download/verify, `latest.yml`, artifact upload, release upload).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -44,9 +44,15 @@ jobs:
       - name: Build renderer
         run: npx vite build
 
-      # ── Package (unpacked directory, no installer) ─────────
+      # ── Package (NSIS installer) ───────────────────────────
+      # Build the NSIS installer so electron-updater can actually install the
+      # update on Windows. The previous `--dir` build shipped a zip, which
+      # electron-updater's NsisUpdater can download and verify but has no way to
+      # install — quitAndInstall() no-ops and the app relaunches unchanged
+      # (issue #96). --publish never: we upload manually below so latest.yml is
+      # generated AFTER signing (see the note on that step).
       - name: Package with electron-builder
-        run: npx electron-builder --win --dir --publish never
+        run: npx electron-builder --win --publish never
 
       - name: Extract version
         id: version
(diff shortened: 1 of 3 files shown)


amirlehmam:
Merged and shipped in [v0.26.0](https://github.com/amirlehmam/wmux/releases/tag/v0.26.0) — thank you, the analysis in this PR (and #96) was spot-on, including the reasoning for gating `latest.yml` on signing.

Two follow-up commits on master after the merge:
- `5a2be62` restores the portable zip as an *additional* release asset — the docs and wmux.org point users at `wmux-X.Y.Z-win-x64.zip`, and it's the SmartScreen-friendliest artifact. `latest.yml` still points only at the installer, exactly as you set it up.
- `120edd5` — the first v0.26.0 run surfaced that the SignPath project currently signs with a **self-signed** cert (`CN=Amir Lehmam`), which `Get-AuthenticodeSignature` reports as `UnknownError` and your (correctly) fatal verify step killed the whole release. The verify step now marks the artifact trusted/untrusted instead of failing, ships it either way, and only publishes `latest.yml` when chain-trusted — same invariant you designed, just tolerant of the self-signed case.

PR #101, opened 2026-07-17
fix(orchestration): write the status vocabulary the app counts (fixes #99)

### LINKED ISSUE #99: Degraded-mode runs stick at 0/N: orchestration hooks write "completed", the app counts "exited"/"complete"
## Summary

In degraded mode (Claude Code native subagents, where the `SubagentStop` hook drives wave transitions), the orchestration cockpit stays stuck at **0/N agents complete** and waves never show as finished — even though every agent ran successfully and the run actually progresses.

The cause is a status vocabulary mismatch: `on-agent-stop.sh` writes `completed`, but the app's canonical vocabulary is `exited` for agents and `complete` for waves/run.

## Evidence

The canonical vocabulary is defined in `src/shared/types.ts:348-349`:

```ts
export type OrchAgentStatus = 'pending' | 'running' | 'exited' | 'failed';
export type OrchWaveStatus  = 'pending' | 'running' | 'complete' | 'failed';
export type OrchRunStatus   = 'pending' | 'running' | 'complete' | 'failed';
```

`completed` is not a member of any of them. Consumers agree with the types:

- `src/renderer/components/Sidebar/OrchestrationPanel.tsx:39` — `countByStatus(orch, 'exited')`
- `OrchestrationPanel.tsx:107` — `a.status === 'exited' || a.status === 'failed'`
- `resources/wmux-orchestrator/scripts/dashboard-text.js` — reads `exited` (96, 98, 128, 173, 207) and `complete` (104, 111, 180)

And the orchestrate skill already documents this explicitly — `skills/orchestrate/SKILL.md:446`:

> **vocabulary the sidebar actually counts** (`exited` for agents, `complete` for waves/run — NOT `completed`)

with the correct call shown at `SKILL.md:452`:

```bash
update_agent "[orch-dir]" "<id>" "status=exited" "exitCode=0" "finishedAt=$NOW"
```

So the skill doc warns humans about the trap, but the script itself still falls into it.

## Affected sites

`resources/wmux-orchestrator/scripts/on-agent-stop.sh` — writes the wrong words:

| Line | Writes | Should be |
|---|---|---|
| 22 | `status=completed` (agent) | `status=exited` |
| 30 | `.waves[N].status` = `completed` | `complete` |
| 40 | `.status` = `completed` (run) | `complete` |

## Note for whoever fixes this — the two files are coupled

`json-tool.js` currently *reads* `completed`, which makes it self-consistent with the broken script:

- `json-tool.js:208` — `agents.every(a => a.status === 'completed' || a.status === 'failed')` (backs `wave_complete()`)
- `json-tool.js:332` — `if (agent.status === 'completed') completedAgents++;` (markdown dashboard counter)

**Fixing `on-agent-stop.sh` alone would break wave transitions entirely** — `wave_complete()` would stop returning true and waves would never advance, which is worse than the current display bug. Both files have to move to `exited`/`complete` together.

Note that `dashboard-text.js` already migrated to the correct vocabulary; `on-agent-stop.sh` + `json-tool.js` appear to be the only remaining holdouts.

## Reproduction

Run an orchestration in degraded mode (native subagents rather than `wmux agent spawn`) so the `SubagentStop` hook fires. Agents complete successfully, but the sidebar's OrchestrationPanel keeps reporting 0/N and the wave never renders as complete.

## Environment

- wmux v0.26.0 (`120edd5`)
- Platform-independent — not OS-specific. `hooks/hooks.json` invokes `bash "${CLAUDE_PLUGIN_ROOT}/scripts/on-agent-stop.sh"` unconditionally with no platform branch, and there are no `.ps1`/`.cmd` variants of these scripts, so Windows executes the identical file. The mismatch is a static property of the source rather than anything runtime.

I have a fix with test coverage ready and am happy to open a PR if you'd like.

ConnorRobinson13:
Fixes #99.

In degraded mode (native subagents, where the `SubagentStop` hook drives wave transitions), `on-agent-stop.sh` marked finished agents `completed` and finished waves/runs `completed`. Neither word exists in the app's vocabulary — `src/shared/types.ts:348-349` defines agents as `exited` and waves/runs as `complete`, and the sidebar counts `exited` (`OrchestrationPanel.tsx:39`). So the cockpit sat at 0/N forever even though every agent had finished.

`skills/orchestrate/SKILL.md:446` already documents the correct vocabulary and explicitly warns against `completed`, and `dashboard-text.js` had already migrated — the script just never followed its own doc.

## Why this touches `json-tool.js` too

`json-tool.js` read `completed` back in two places, which made it self-consistent with the broken script:

- `:208` — `wave-complete`, which backs `wave_complete()`
- `:332` — the markdown dashboard counter

Fixing only the shell script would have left `wave-complete` never returning true, stopping wave transitions altogether — worse than the display bug. Both files have to move together, so they're in one commit.

## Changes

| File | Change |
|---|---|
| `on-agent-stop.sh:22` | agent `completed` → `exited` |
| `on-agent-stop.sh:30` | wave `completed` → `complete` |
| `on-agent-stop.sh:40` | run `completed` → `complete` |
| `json-tool.js:208` | read `exited` |
| `json-tool.js:332` | read `exited` |

## Tests

`tests/unit/orchestration-status-vocab.test.ts` runs the real `on-agent-stop.sh` end-to-end against a temp orchestration dir (via `TMPDIR`, the same discovery path `find_active_orch()` uses) and asserts the resulting `state.json`. The expected status arrays are typed against `OrchAgentStatus` / `OrchWaveStatus` / `OrchRunStatus`, so they stop compiling if the unions change — the shell and TypeScript layers can't drift apart silently again. There's also a guard test that no script reintroduces a `status=completed` write.

7 of the 9 tests fail on `master` and all 9 pass with this change. Full suite is green.

## One thing worth flagging

`CLAUDE.md` notes the orchestrator plugin is also published standalone at `amirlehmam/wmux-orchestrator`. The same mismatch is presumably there, and this PR doesn't reach it — flagging in case it needs a companion fix.

Happy to adjust anything here, including splitting the tests out or trimming them down if they're heavier than you'd like for this.

### DIFF
--- a/resources/wmux-orchestrator/scripts/json-tool.js
+++ b/resources/wmux-orchestrator/scripts/json-tool.js
@@ -205,7 +205,7 @@ function cmdQuery(file, queryName, ...args) {
         break;
       }
       const agents = data.waves[waveIdx].agents || [];
-      const allDone = agents.every(a => a.status === 'completed' || a.status === 'failed');
+      const allDone = agents.every(a => a.status === 'exited' || a.status === 'failed');
       process.stdout.write(allDone ? 'true' : 'false');
       process.stdout.write('\n');
       break;
@@ -329,7 +329,7 @@ function cmdDashboard(file) {
   for (const wave of waves) {
     for (const agent of (wave.agents || [])) {
       totalAgents++;
-      if (agent.status === 'completed') completedAgents++;
+      if (agent.status === 'exited') completedAgents++;
       else if (agent.status === 'running') runningAgents++;
       else if (agent.status === 'failed') failedAgents++;
     }
--- a/resources/wmux-orchestrator/scripts/on-agent-stop.sh
+++ b/resources/wmux-orchestrator/scripts/on-agent-stop.sh
@@ -19,7 +19,7 @@ EXIT_CODE="${CLAUDE_EXIT_CODE:-0}"
 NOW=$(date -u +%Y-%m-%dT%H:%M:%SZ)
 
 if [ "$EXIT_CODE" = "0" ]; then
-  update_agent "$ORCH_DIR" "$AGENT_ID" "status=completed" "exitCode=0" "finishedAt=$NOW"
+  update_agent "$ORCH_DIR" "$AGENT_ID" "status=exited" "exitCode=0" "finishedAt=$NOW"
 else
   update_agent "$ORCH_DIR" "$AGENT_ID" "status=failed" "exitCode=$EXIT_CODE" "finishedAt=$NOW"
 fi
@@ -27,7 +27,7 @@ fi
 WAVE_IDX=$(node "$JSON_TOOL" query "$ORCH_DIR/state.json" wave-of-agent "$AGENT_ID" 2>/dev/null)
 
 if [ -n "$WAVE_IDX" ] && wave_complete "$ORCH_DIR" "$WAVE_IDX"; then
-  update_state "$ORCH_DIR" ".waves[$WAVE_IDX].status" "completed"
+  update_state "$ORCH_DIR" ".waves[$WAVE_IDX].status" "complete"
 
   if all_waves_done "$ORCH_DIR"; then
     REVIEWER_STATUS=$(read_state "$ORCH_DIR" '.reviewer.status')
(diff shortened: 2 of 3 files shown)


amirlehmam:
Merged — thank you for a model PR. Verified the vocabulary against src/shared/types.ts and OrchestrationPanel locally, full suite green (223/223). Really like the tests being typed against the OrchAgentStatus/OrchWaveStatus unions so shell/TS drift becomes a compile error, and the guard test against reintroducing status=completed. Your flag about the standalone repo was right: amirlehmam/wmux-orchestrator had the same bug and is now fixed (c1f7eef) with the plugin bumped to 0.1.2 so cached installs refresh. Ships in v0.27.0.

PR #103, opened 2026-07-18
Add right-click context menu for terminal tabs

dhyantsoni:
Right-clicking a terminal tab now opens a context menu with three actions:

- **Rename** – rename the tab inline
- **Close** – close the tab
- **Close others** – close every other tab in the pane (shown only when more than one tab is open)

This matches the context menu that sidebar workspaces already have. Added a `closeOtherSurfaces` store action with unit tests.


amirlehmam:
Merged — thanks Dhyan! Nice touch reusing the sidebar's `ctx-menu` styling so the tab menu matches visually, and `closeOtherSurfaces` correctly mirrors `closeSurface`'s bookkeeping (closed-tab history for Ctrl+Shift+T + PTY reaping), so nothing leaks. Shipping in v0.27.2.

PR #104, opened 2026-07-18
Add rename-surface CLI command

dhyantsoni:
The CLI can rename workspaces (`rename-workspace`) but there was no equivalent for tabs. This adds a `surface.rename` pipe method and a `rename-surface` CLI command.

```
wmux rename-surface <surfaceId> <title>
wmux rename-surface <title>            # renames the current surface when run inside a pane
```

Follows the existing table-driven pipe-bridge pattern used by the other surface commands.


amirlehmam:
Merged — thanks! This filled a real gap in the CLI (workspaces could be renamed but tabs couldn't), and the `WMUX_SURFACE_ID` fallback makes `wmux rename-surface <title>` work naturally from inside a pane, which is exactly how agents will use it. Follows the v2-bridge table pattern cleanly. Shipping in v0.27.2.

PR #105, opened 2026-07-18
Add duplicate tab option

dhyantsoni:
Adds a **Duplicate tab** item to the new-tab (`+`) dropdown.

It opens a copy of the current tab in the same pane, reusing its type, shell, color scheme and current working directory, so you get a fresh shell right where you already are. Startup commands are not replayed, so duplicating an agent tab won't relaunch the agent.

Added a `duplicateSurface` store action with unit tests.


amirlehmam:
Merged — thanks! Good call preferring `currentCwd` over the original `cwd` (duplicating opens where the shell actually is now, not where it started) and intentionally not replaying startup commands so duplicating an agent tab doesn't relaunch the agent. Note: I resolved the merge conflicts with #103/#106 on your branch (both landed first and touch the same tab-bar files) — all three features coexist in the merged tree with 244 tests green. Shipping in v0.27.2.

PR #106, opened 2026-07-18
Add close tabs to the right

dhyantsoni:
Adds a **Close to the right** item to the tab context menu. It closes every tab positioned after the one you clicked, and only shows when there are tabs to the right.

Added a `closeSurfacesToRight` store action with unit tests.

Note: this builds on the tab context menu from #103, so it's best merged after that PR.


amirlehmam:
Merged — thanks! Stacking this on #103 made the merge sequence trivial, the item correctly hides when there's nothing to the right, and clamping `activeSurfaceIndex` covers the case where the active tab was one of the closed ones. Shipping in v0.27.2.

PR #108, opened 2026-07-19
Show the file name on markdown tabs instead of a generic "Markdown" label

AvoChang:
## Problem

When a markdown file is opened in a tab, the tab always reads **"Markdown"** regardless of which document it holds. With several markdown tabs open in a pane they're indistinguishable — you can't tell `README.md` from `CHANGELOG.md` without clicking through them.

## Change

Markdown tabs now show the backing file's **basename** (e.g. `README.md`) as their label. When a surface has no file behind it — content pushed via `wmux markdown set --content` — it keeps the generic `Markdown` label. An explicit custom title still wins.

### How it works

- Add `markdownFileName?: string` to `SurfaceRef`.
- Thread an optional `fileName` through `setMarkdownContent` and the `__wmux_setMarkdownContent` renderer bridge. It's populated from the file basename at every file-loading entry point:
  - the `markdown.load_file` pipe handler — covers `wmux markdown <file>` and `wmux markdown set <id> --file <path>`
  - the command-palette "Open Markdown File" picker
- Content-only setters (`markdown.set_content`) don't pass a name, so they leave the label untouched.
- `getSurfaceLabel` returns `markdownFileName` for markdown surfaces, still preferring an explicit `customTitle`.

## Tests

- `tests/unit/surface-label.test.ts`: file name shown, fallback to "Markdown", custom title precedence.
- `tests/unit/markdown-content.test.ts`: file name recorded when provided, left untouched on a content-only update.

All affected unit tests pass; lint clean; `npm run build:main` compiles. (The pre-existing `pty-manager` test failure is an unrelated node-pty native-spawn issue on Windows and does not touch this change.)

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -606,7 +606,7 @@ app.whenReady().then(() => {
             const win = BrowserWindow.getAllWindows()[0];
             if (!win || win.isDestroyed()) { respondError(-32000, 'No window'); return; }
             await win.webContents.executeJavaScript(
-              `window.__wmux_setMarkdownContent?.(${JSON.stringify(request.params?.surfaceId || '')}, ${JSON.stringify(content)})`
+              `window.__wmux_setMarkdownContent?.(${JSON.stringify(request.params?.surfaceId || '')}, ${JSON.stringify(content)}, ${JSON.stringify(path.basename(filePath))})`
             );
             respond({ ok: true, length: content.length });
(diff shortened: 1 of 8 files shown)


amirlehmam:
Merged — thanks @AvoChang! Clean implementation: threading the basename through every file-loading entry point while leaving `markdown.set_content` untouched is exactly the right split, and the tests covering the custom-title precedence are appreciated. Shipping in v0.28.0.

PR #110, opened 2026-07-19
fix(pipe): report the real app version from system.identify

ConnorRobinson13:
## Problem

`system.identify` returns a hardcoded `version: '0.5.0'` (`src/main/index.ts`), which is ~22 releases stale against the current `0.27.2`. Any client that keys on the handshake version — instance detection, and cmux-ecosystem tooling that reads `system.identify` — sees the wrong number.

## Fix

Return `app.getVersion()` — already the source of truth used elsewhere in the same file (e.g. `handleVersionChange(app.getVersion())` at `index.ts:335`, and the `SYSTEM_GET_VERSION` IPC handler). One line, no new imports (`app` is already imported).

`npm run build:main` clean.

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -398,7 +398,7 @@ app.whenReady().then(() => {
 
     switch (request.method) {
       case 'system.identify':
-        respond({ name: 'wmux', version: '0.5.0', platform: 'win32' });
+        respond({ name: 'wmux', version: app.getVersion(), platform: 'win32' });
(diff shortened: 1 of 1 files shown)


amirlehmam:
Merged — thanks @ConnorRobinson13! The hardcoded `0.5.0` had drifted ~22 releases; `app.getVersion()` is the right source of truth. Shipping in v0.28.0.

Issue #64, opened 2026-06-26
feat: add a few high-value keyboard shortcuts + shortcut cheat-sheet/tooltips


kevmtt:
## Motivation

wmux already ships a strong, remappable shortcut set (38 actions, listed in the `Ctrl+Shift+P` command palette, editable in **Settings → Keyboard**) — all the workspace/tab/split basics are done. This issue proposes a small set of **high-productivity** shortcuts that are still missing vs [cmux](https://cmux.com/docs/keyboard-shortcuts), plus a discoverability feature. Kept intentionally focused to avoid bloating the shortcut surface.

## Already covered (no work needed)

| Capability | wmux action | wmux default | cmux equivalent |
|---|---|---|---|
| New workspace | `newWorkspace` | `Ctrl+N` | `⌘N` |
| Switch workspace (cycle / index) | `nextWorkspace` / `prevWorkspace` / built-in | `Ctrl+PgDn` / `Ctrl+PgUp` / `Ctrl+1…9` | `⌃⌘]` / `⌃⌘[` / `⌘1…9` |
| Jump to workspace with unread notification | `jumpToUnread` | `Ctrl+Shift+U` | `⌘⇧U` |
| Rename workspace | `renameWorkspace` | `Ctrl+Shift+F2` | `⌘⇧R` |
| New terminal tab | `newSurface` | `Ctrl+T` | `⌘T` |
| Next / previous tab | `nextSurface` / `prevSurface` | `Ctrl+Shift+]` / `Ctrl+Shift+[` | `⌘⇧]` / `⌘⇧[` |
| Split right / down | `splitRight` / `splitDown` | `Ctrl+D` / `Ctrl+Shift+D` | `⌘D` / `⌘⇧D` |

> The five shortcuts originally requested (new workspace, switch workspace, switch-to-notification, new tab, split) are all in this table — so this issue only adds genuinely new, high-value actions.

## Proposed additions (high-value only)

Suggested Windows defaults are collision-checked against current defaults and remappable.

| Proposed action | What it does | Why it's worth it | Suggested default | cmux |
|---|---|---|---|---|
| `selectSurface1…9` | Jump straight to tab N in the focused pane | Fast tab switching when juggling several agents/shells | `Ctrl+Alt+1…9` | `⌃1–9` |
| `reopenClosedSurface` | Restore the last-closed tab | Universal "oops, brought it back" muscle memory | `Ctrl+Shift+T` | `⌘⇧T` |
[...]

PR #111, opened 2026-07-21
fix(terminal): Shift+Enter inserts a newline in TUI apps (Claude Code)

AvoChang:
## Problem

In a wmux terminal running Claude Code, **Shift+Enter does not insert a newline** — it submits the prompt instead. This works in cmux but was broken in wmux.

## Root cause

xterm.js emits a plain `\r` (carriage return) for **both** Enter and Shift+Enter, so a TUI app can't distinguish them and treats Shift+Enter as submit. wmux's custom key handler in `useTerminal.ts` had no mapping for Shift+Enter.

Claude Code recognizes `ESC + CR` (`\x1b\r`) as a newline — the same sequence Alt/Option+Enter produces, and exactly what Claude Code's `/terminal-setup` configures for iTerm2/VSCode.

## Fix

Intercept Shift+Enter in the terminal's `attachCustomKeyEventHandler` and write `\x1b\r` to the PTY, returning `false` to suppress xterm's default `\r`.

- Excludes Ctrl/Alt/Meta, so **Ctrl+Shift+Enter (zoom pane)** and Alt+Enter are left untouched.
- No effect on plain Enter (submit) behavior.

## Testing

- `npm run build:renderer` passes.
- Manual: in Claude Code, Shift+Enter now inserts a newline; Enter still submits; Ctrl+Shift+Enter still zooms the pane.

### DIFF
--- a/src/renderer/hooks/useTerminal.ts
+++ b/src/renderer/hooks/useTerminal.ts
@@ -692,6 +692,25 @@ export function useTerminal({ surfaceId, shell, cwd, visible = true, focused = t
         })();
         return false; // Prevent default — we handle paste ourselves
       }
+      // Shift+Enter → newline for TUI apps (Claude Code, etc). xterm sends a
+      // plain \r for BOTH Enter and Shift+Enter, so the app can't tell them
+      // apart and treats Shift+Enter as submit. Emit ESC+CR — the same sequence
+      // Alt/Option+Enter produces and that Claude Code's /terminal-setup wires
+      // up — so Shift+Enter inserts a newline instead of submitting. Excludes
+      // Ctrl/Alt/Meta so Ctrl+Shift+Enter (zoom pane) is left untouched.
+      if (
+        event.type === 'keydown' &&
+        event.key === 'Enter' &&
+        event.shiftKey &&
+        !event.ctrlKey &&
+        !event.altKey &&
+        !event.metaKey
+      ) {
+        if (ptyIdRef.current) {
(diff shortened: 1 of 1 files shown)


arnonuem:
i use another harness than claude code and for me shift enter works (windows)

amirlehmam:
Merged — thanks @AvoChang, this is a clean fix and the choice of `ESC CR` is exactly right: it's the same sequence Alt+Enter produces and the one Claude Code's `/terminal-setup` wires up in other terminals, so TUI input parsers already treat it as "insert newline". The modifier guards keeping Ctrl+Shift+Enter (zoom pane) untouched were appreciated too.

One small follow-up landed on master (2b7202f): the sequence is now sent via `terminal.input()` instead of `pty.write()` directly, so it flows through `onData` and respects broadcast-input mode (#64) — otherwise Shift+Enter would only reach the focused pane while every other keystroke fans out to all panes.

@arnonuem re: "shift enter works for me in another harness" — worth noting that xterm.js (which wmux embeds) doesn't support `modifyOtherKeys` or the kitty keyboard protocol, so before this PR *every* app in wmux received a plain `\r` for both Enter and Shift+Enter; nothing could actually distinguish them. If your harness appeared to handle Shift+Enter, it was treating plain `\r` contextually. After this change it will receive `ESC CR` (the standard Alt+Enter encoding) instead — virtually all terminal input parsers handle that atomically as "newline", but if you see any regression in your harness after v0.29.0, please open an issue and we'll gate it.

Shipping in v0.29.0.

PR #115, opened 2026-07-23
feat(browser): configurable dev-server ports + fix auto-open selection

ConnorRobinson13:
## What

Adds a `[browser]` section to `~/.wmux/config.toml` so users can extend the set of dev-server ports that trigger browser auto-navigation, and toggle the auto-open behavior:

```toml
[browser]
dev-ports = [8501, 4321]   # merged with the built-in defaults, de-duped
auto-open = true           # set false to disable browser auto-navigation
```

### Why

The recognized dev-port set was a hardcoded array in `App.tsx`, so a server on any other port (Streamlit's 8501, FastAPI on 8001, etc.) was silently never picked up, with no way to change it short of editing source.

- Ports are **additive** — configured ports join the built-in defaults rather than replacing them — so adding one never drops the sensible defaults.
- `auto-open = false` keeps port detection but stops the browser from auto-navigating, for users who find it disruptive.

Follows the existing `~/.wmux/config.toml` pattern (`[terminal]`, `[appearance]`): same kebab/camelCase key aliasing and non-fatal validation warnings.

## Also fixes a pre-existing auto-open bug

While testing on a machine with several dev servers already running, the browser never opened the newly-started one. Root cause in `handlePortsUpdate`: it always navigated to `matchDevPorts(...)[0]` — whichever recognized port `netstat` happened to list first — then wrote the full recognized set into `ws.ports`, so the "is this a new port?" guard permanently suppressed navigation afterward. Net effect: it opened the first-seen dev port once, then went silent.

It now opens the first recognized port the workspace hasn't seen yet, so a freshly-started server opens even when other dev ports are already listening.

## Changes

- `src/main/user-config.ts` — parse the `[browser]` section (`dev-ports` validated to 1–65535 integers, `auto-open` bool).
- `src/renderer/dev-ports.ts` *(new)* — pure helpers `mergeDevPorts` / `matchDevPorts` / `firstNewDevPort`, extracted so the logic is unit-testable.
- `src/renderer/App.tsx` — config drives the active port set; auto-open targets the newly-appeared port.
- Tests — `tests/unit/dev-ports.test.ts` *(new, 11 cases)* + the `[browser]` cases in `tests/unit/user-config.test.ts`.

## Testing

- `npm test` green (19 assertions across the two files; full suite passes).
- Verified live on Windows with a real dev server: config-driven detection of a custom port (8501) and the auto-open fix both confirmed, including the busy-machine scenario the fix targets.

### DIFF
--- a/src/main/user-config.ts
+++ b/src/main/user-config.ts
@@ -27,6 +27,10 @@
  *   [appearance]
  *   ui-theme = "light"   # light | dark | system (issue #67)
  *
+ *   [browser]
+ *   dev-ports = [8501, 4321]   # extra dev-server ports, merged with built-in defaults
+ *   auto-open = true           # auto-navigate the browser to a newly-detected dev port
+ *
  * File-wins-at-startup, app-wins-at-runtime: this data seeds the store
  * on boot; users can still tweak via the Settings UI afterwards.
  * A `wmux reload-config` command re-applies the file over runtime state.
@@ -61,6 +65,13 @@ export interface UserConfig {
   /** App UI theme (issue #67) — separate from the terminal color scheme. */
   appearance?: {
     uiTheme?: UiTheme;
+  };
+  /** Browser surface behavior — dev-server port detection & auto-navigation. */
+  browser?: {
+    /** Extra ports (merged with the built-in defaults) that count as dev servers. */
+    devPorts?: number[];
+    /** Auto-navigate the workspace browser to a newly-detected dev port (default true). */
+    autoOpen?: boolean;
   };
   /** Absolute path the config was read from (for diagnostics). */
   path?: string;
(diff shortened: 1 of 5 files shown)


amirlehmam:
Merged — thanks for a really well-put-together PR. The additive port merge, the `auto-open` toggle, and especially catching the pre-existing `devPorts[0]` selection bug (with a regression test for the busy-machine case) made this an easy review; extracting the pure helpers into `dev-ports.ts` is exactly the pattern we want.

One small follow-up landed on master (73b4e2b): `applyUserConfigBrowser` now resets to the built-in defaults before applying the config, so `wmux reload-config` also *reverts* the effect of a key you've removed from `config.toml` instead of keeping the previous run's values sticky until restart.

Shipping in **v0.33.0** (CI is building it now).

Issue #116, opened 2026-07-24
Markdown viewer: path awareness, copy, source/preview toggle, and in-place editing


kevmtt:
# Markdown viewer: copy, source/preview toggle, and in-place editing

## Summary

The markdown surface is currently a **one-way, read-only render target**: content
goes in via CLI/pipe/file-picker, gets converted to sanitized HTML, and that's it.
There is no way to get the text back out (other than mouse-selecting rendered
HTML, which loses markdown syntax), no way to look at the raw source, and no way
to fix a typo in the file you're reading.

That makes the viewer good for *consuming* agent output (its original purpose,
issue #54) but useless for the far more common loop: **agent writes a plan/spec →
user reads it in the pane → user wants to copy it into a ticket, quote one line
into chat, or tweak two sentences and save.**

It also means the surface forgets *where* the content came from: the full path is
dropped at every entry point, so the pane can't tell you which file you're
reading, can't hand the path to the terminal next door, and can't save back to it.

This issue proposes four features, in priority order:

| # | Feature | Priority | Rough size |
|---|---------|----------|-----------|
| F0 | **Path-aware surfaces** — a markdown surface knows, shows, and can hand out the path of the file it came from | must have | small |
| F1 | **Copy Markdown** — copy the raw markdown source to the clipboard (all, a selection, or a code block) — *and the file path* | must have | small |
| F2 | **Toggle Text/Markdown view** — switch the pane between rendered preview and raw, line-selectable source | must have | small |
| F3 | **Edit & Save** — edit the source in place and write it back to the backing file | nice to have | medium |

F0–F2 are pure-renderer changes (F0 needs one field plumbed through existing
producers) with no new IPC and no new trust surface. F3 introduces the first
**write** path to the filesystem from the markdown surface and needs a deliberate
security design (see [Security](#security)).

---

### Rendering

`src/renderer/components/Markdown/MarkdownPane.tsx:12-51` is the whole component
(51 lines). It:

1. `marked.parse(content)` with `{ gfm: true, breaks: true }`
2. `DOMPurify.sanitize(...)` with `FORBID_TAGS: ['style','form','input','button','textarea','select']`
3. dumps the result through `dangerouslySetInnerHTML` into `.markdown-pane__content`
4. intercepts clicks on `<a>` and routes them to the wmux browser panel

No toolbar, no state, no modes. The raw `content` prop is available in the
component but never shown to the user.

### Where content comes from

[...]

PR #120, opened 2026-07-27
refactor(i18n): one file per locale with keys typed against English; add Italian

simplyBarbe:
Adding a language meant editing **five** places, and the fifth was not in the i18n folder at all:

| # | Location | What breaks if you forget |
|---|---|---|
| 1 | `core.ts` — the `Language` union | won't compile (good) |
| 2 | `core.ts` — the `LANGUAGES` array | missing from the Settings dropdown |
| 3 | `core.ts` — the dictionary const | — |
| 4 | `core.ts` — the `DICTS` map | won't compile (good) |
| 5 | **`settings-slice.ts`** — `candidate === 'en' \|\| 'fr' \|\| 'zh'` | **silent**: the language persists, then resets to the detected default on next launch |

That last one is the dangerous kind — no compile error, no failing test, just a bug discovered after shipping.

Two other problems: `Record<string, string>` meant nothing was checked, so `t('setings.title')` compiled and rendered the raw key; and at ~90 lines per language a single 351-line file would put every translation PR in conflict with every other.

### What changed

Dictionaries move out of `core.ts` into one file per locale. English is the source of truth and exports the key type:

```ts
// locales/en.ts
export const en = { 'settings.title': 'Settings', /* … */ } as const;
export type TranslationKey = keyof typeof en;
export type Translation = Partial<Record<TranslationKey, string>>;
```

`Partial` is the deliberate choice: a **missing** key is legal (it falls back to English, so partial translations still ship), but an **unknown** key is a compile error — that catches typos and keys stranded by an English rename. `useT()` and `translate()` now take `TranslationKey`, extending the same check to all 79 call sites.

`core.ts` keeps a single table everything derives from:

```ts
const REGISTRY = [
  { code: 'en', label: 'English',  dict: en as Translation },
  { code: 'fr', label: 'Français', dict: fr },
  { code: 'it', label: 'Italiano', dict: it },
  { code: 'zh', label: '中文',      dict: zh },
] as const;

export type Language = (typeof REGISTRY)[number]['code'];
```

`Language`, `SUPPORTED_LANGUAGES`, `LANGUAGES` and the new `isLanguage()` guard all derive from it, and `settings-slice.ts` uses `isLanguage()` instead of the hardcoded list — closing case #5.

Also fixed: `SettingsWindow`'s `TAB_LABEL_KEYS` was annotated `Record<Tab, string>`, which widened the literals away. Typing it as `TranslationKey` checks those nine keys where they are written. This one was found by the compiler, not by review — exactly the drift the refactor exists to catch.

**Adding a language is now two edits**: a `locales/xx.ts` file and one row in `REGISTRY`.

### Italian

Added as the first language under the new structure, as a test of that claim — it took the promised two edits. A third file changed because the test pinning the shipped language set failed on its own:

```
AssertionError: expected [ 'en', 'fr', 'it', 'zh' ] to deeply equal [ 'en', 'fr', 'zh' ]
```

That guard is working as designed: the shipped set cannot grow unnoticed. Everything else — the Settings dropdown, `isLanguage()`, the stale-key test — picked up `it` on its own.

**Behavior change:** `detectDefaultLanguage()` now resolves `it-IT`, so on an Italian-display Windows a fresh profile starts in Italian instead of English (the issue #56 case, now covering Italian). Existing installs keep whatever they persisted.

### Contract preserved

`LANGUAGES` still yields exactly `{ code, label }` (asserted — the registry's `dict` field does not leak), and `translate`, `detectDefaultLanguage`, `useT`, `Language` and the `'../i18n/core'` import path are unchanged.

### Verification

- Dictionaries verified **byte-identical** to the previous `core.ts` — old and new transpiled and diffed object-for-object, 84 keys × 3 languages, zero differences.
- Contract proven negatively: a scratch file with three bad cases (unknown key in a locale, typo at a call site, unshipped language code) produced **exactly three** errors, while the three valid cases compiled clean.
- `npm run typecheck` → 0
- `npm test` → **430/430** across 46 files
- `npm run lint` → clean on every touched file (32 pre-existing problems repo-wide, unchanged vs. base)
- Coverage: `en 84/84 · fr 84/84 · it 84/84 · zh 84/84`

### Test changes

The suite gains a stale-key guard per language, a coverage report, `isLanguage` cases, and a cast in the fallback-chain test — that key is now provably invalid, so it needs `as TranslationKey` to still exercise the runtime chain. Worth knowing: vitest strips types, so that one would have stayed green while `typecheck` failed.


amirlehmam:
Merged in 1696aa9 — thank you, this is a good refactor and the reasoning in the description holds up.

## What I verified before merging

The claim the whole PR rests on is that the dictionaries came across untouched, so I checked it independently rather than taking the diff's word for it: extracted the three dicts from the old `core.ts` and the three new `locales/*.ts`, evaluated both, and compared object-for-object. **84 keys each in en/fr/zh, zero value differences.** Beyond that: 449 tests pass, `typecheck` is clean on both projects, and lint is clean on every touched file (the 32 repo-wide problems are pre-existing and unchanged against base).

Case #5 in your table is the one that made this worth doing. A `candidate === 'en' || 'fr' || 'zh'` sitting in `settings-slice.ts` is exactly the kind of drift that ships: no compile error, no failing test, and the symptom (language resets on next launch) shows up far enough from the cause that nobody connects the two. Deriving `isLanguage()` from the registry is the right shape — the guard can't fall behind the thing it guards.

`Partial<Record<TranslationKey, string>>` is also the right call over a full `Record`. A missing key falling back to English is what lets a partial translation ship at all; an *unknown* key being an error is what catches the typo. Getting both from one type is neat.

## One follow-up applied on merge

The Italian strings used French-style spaced guillemets — `« Apri in wmux »`. Italian sets them tight: `«Apri in wmux»`. Four strings, fixed in the merge commit.

## It earned itself within the hour

I shipped issue #116's F3 (markdown edit & save) straight after merging this, which needed 15 new i18n keys. Every single one was a compile error until it was declared in `en.ts` — which is precisely the failure mode you built this to catch, and it caught them before they could render as raw keys in the UI.

It also surfaced a hole I'd have shipped otherwise: I'd typed a `t` parameter as `(key: any, fallback?: string) => string` when threading the translator into a helper, which silently exempted that whole table of labels from the check. Typing it `TranslationKey` turned it back on. Worth knowing as a general hazard with this design — the contract is only as strong as the narrowest `t` in the call chain.

Italian ships in **v0.37.0**, out now.

PR #127, opened 2026-07-29
fix: stop notifying on every shell prompt redraw (ConEmu OSC 9 subcommands)

AxelM-MSK:
## Problem

wmux fires a notification every time a command is run at a shell prompt — extremely noisy alongside the (desired) agent notifications.

## Root cause

The OSC 9 handler in `useTerminal.ts` treats **every** OSC 9 payload as an iTerm2-style notification (`OSC 9;<text>`). But OSC 9 is overloaded: ConEmu/Windows Terminal use numeric subcommands — `OSC 9;9;<cwd>` (cwd report) and `OSC 9;4;<state>;<n>` (progress). wmux's own `wmux-cmd-integration.cmd` embeds `9;9;$P` in the cmd prompt, and the standard Windows Terminal PowerShell profile snippet does the same — so every prompt redraw after every command raised a bogus notification with text like `9;C:\path`.

## Fix

One guard in the OSC 9 handler: payloads starting with a numeric subcommand (`/^\d+;/`) are swallowed instead of raised. Bare-text notifications, `wmux notify`, and the Claude Code Stop/Notification hook path are unaffected.

## Testing

Verified in a dev instance:
- `ESC ]9;9;C:\fake BEL` and `ESC ]9;4;1;50 BEL` → silent (previously each fired a notification)
- `ESC ]9;hello BEL` → still notifies
- cmd tab: pressing Enter repeatedly no longer fires notifications
- `wmux notify` and agent Stop-hook notifications still work

### DIFF
--- a/src/renderer/hooks/useTerminal.ts
+++ b/src/renderer/hooks/useTerminal.ts
@@ -566,6 +566,11 @@ export function useTerminal({ surfaceId, shell, cwd, visible = true, focused = t
     // Register OSC notification handlers
     // OSC 9: basic notification (iTerm2 style)
     terminal.parser.registerOscHandler(9, (data) => {
+      // ConEmu/Windows Terminal overload OSC 9 with numeric subcommands:
+      // "9;<cwd>" (cwd report — our own cmd integration and the standard WT
+      // PowerShell prompt snippet emit this on EVERY prompt redraw) and
+      // "4;<state>;<n>" (progress). Only bare text is an iTerm2 notification.
+      if (/^\d+;/.test(data)) return true;
       window.wmux.notification.fire({
         surfaceId: ptyIdRef.current || '',
         text: data,


amirlehmam:
Merged, thank you — and the diagnosis turned out to uncover a second, older bug.

Your root-cause analysis was exactly right: `wmux-cmd-integration.cmd:14` embeds `$e]9;9;$P` in the prompt, so every redraw was raising a notification with text like `9;C:\path`.

While verifying the fix I checked what else registers on OSC 9, and found this: xterm dispatches OSC handlers **newest-registered-first**, stopping at the first one returning `true`. `ProgressAddon` also registers on OSC 9, but it's loaded at `useTerminal.ts:431` — *before* the notification handler at :568. So the notification handler always wins the sequence, and `return true` still consumed `9;4`.

I confirmed it with real xterm + real `ProgressAddon` in wmux's registration order:

| handler returns | notified | ProgressAddon fired |
|---|---|---|
| master (no guard) | `4;1;50`, `9;C:\some\path`, `hello world` | — |
| `true` (this PR) | `hello world` | — |
| `false` | `hello world` | `{state: 1, value: 50}` |

The OSC 9;4 progress bar — tab strip, sidebar, Windows taskbar — has been fed nothing since it shipped in 0.23.0, because this handler has been eating every OSC 9 the whole time.

So I merged your PR as-is and pushed a follow-up (2be9d41) flipping `true` → `false`, which declines the sequence and lets xterm continue down the chain. Your notification fix is unchanged; progress now works for the first time. The predicate moved to `hooks/osc9.ts` with a regression test driving both consumers.

Shipping in v0.39.0.

PR #129, opened 2026-07-29
feat(i18n): add Spanish (Español) locale

alejandrocastrib:
Adds Spanish as a shipped UI language, following the per-locale structure introduced in #120.

That PR closed with a claim: *"Adding a language is now two edits: a `locales/xx.ts` file and one row in `REGISTRY`."* This is the second language added under it, and the claim holds — the third file changed only because a test is designed to fail when the shipped set grows.

| File | Change |
|---|---|
| `src/renderer/i18n/locales/es.ts` | **new** — 107 keys |
| `src/renderer/i18n/core.ts` | +2 lines: the import and `{ code: 'es', label: 'Español', dict: es }` |
| `tests/unit/i18n.test.ts` | the pinned language-set assertion, plus `es` cases |

Everything else derived from `REGISTRY` on its own: the Settings dropdown, `isLanguage()` in `settings-slice.ts` (so the setting persists across restarts — case #5 from #120 stays closed), and the stale-key test, which now runs for `es` without being told to.

`es` is placed after `en` in the registry to keep the table in code order; the Settings dropdown follows that same order.

### Coverage

Complete — nothing falls back to English:

```
  en (English): 107/107 keys
  es (Español): 107/107 keys
  fr (Français): 107/107 keys
  it (Italiano): 107/107 keys
  zh (中文): 107/107 keys
```

Spanish as spoken in Spain, addressing the user informally (*"Haz clic para…"*), matching the direct tone of the English source. `wmux`, `Ctrl+S`, `HKCU`, `markdown set` and the other command/UI literals are left untranslated. Terminology follows the existing Spanish translation on the website (`site/i18n.js`): *barra lateral*, *pestañas*, *navegador integrado*.

### Behavior change

`detectDefaultLanguage()` now resolves `es-*`, so on a Spanish-display Windows a fresh profile starts in Spanish instead of English — the issue #56 case, now covering Spanish. Existing installs keep whatever they persisted.

### Verification

- `npm run typecheck` → 0
- `npm test` → **507 passed**, 5 failed. The 5 failures are in `tests/unit/orchestration-status-vocab.test.ts` and are **pre-existing on `master`**: they shell out to `bash` with a Windows path (`/bin/bash: C:UsersAlejandroCastillo-Ap...: No such file or directory`) and fail identically with these changes stashed. `tests/unit/i18n.test.ts` → **15/15**.
- `npx eslint src/renderer/i18n/` → clean. `npm run lint` → the same 32 repo-wide problems as base, unchanged.

### Test changes

Three edits, all in `tests/unit/i18n.test.ts`:

1. The pinned set becomes `['en', 'es', 'fr', 'it', 'zh']` — the guard from #120 working as designed, since the shipped set cannot grow unnoticed.
2. `translate('es', 'settings.title')` → `'Ajustes'`, alongside the existing `fr`/`zh` cases.
3. `es-ES` added to the OS display-language test, next to `fr-CA` and `zh-Hans-CN`, pinning the regional-tag → base-tag mapping for the new language.


amirlehmam:
Merged, thank you — and thanks for the unusually complete write-up.

Verified locally before merging:

- `es` key parity with `en`: **107/107**, no missing keys, no extras
- `tests/unit/i18n.test.ts` → 15/15
- full suite → 512 passed, `npx tsc --noEmit` clean

On the 5 failures you saw in `orchestration-status-vocab.test.ts`: they're environment-specific, not pre-existing on `master` in general — they pass here (51 files / 511 tests green on the merge-base). They're shelling out to `bash` with a Windows path on your machine, so it's a local Git-Bash resolution issue rather than something in the repo. Nothing for you to fix.

Good call flagging that the pinned language-set assertion is the guard from #120 working as designed — that's exactly why it's pinned.

Shipping in v0.39.0.

PR #131, opened 2026-07-29
perf(sidebar): coalesce the resize drag to one width update per frame

ConnorRobinson13:
One sidebar drag does far more work than it needs to. Not a user-visible break on its own — this is churn, not correctness.

## Cause

`SidebarResizeHandle.tsx` handles `mousemove` raw, with no throttle or rAF coalescing:

```ts
const onMouseMove = (ev: MouseEvent) => {
  const delta = ev.clientX - startX;
  onWidthChange(delta);
};
```

So ~60 width updates a second while dragging. Each one sets `sidebarWidth` in `App.tsx`, which relayouts the workspace — which wakes **every** terminal's `ResizeObserver` (`useTerminal.ts:858`), calls `fit()` and `pty.resize()`, and makes ConPTY re-emit its visible screen. A one-second drag is sixty rounds of that across every open terminal.

`sidebarWidth` is also a dependency of the auto-save listener effect and of `handleSaveSession`, so every intermediate value tears down and re-registers the listener and rebuilds the callback.

## Fix

`createResizeDrag` collects pointer positions and emits at most one delta per frame. `stop()` flushes a position still waiting for its frame, so the width settles where the cursor was released rather than a frame behind it.

The frame scheduler is injected rather than calling `requestAnimationFrame` directly. That's the one design decision worth pointing at: it's what makes the coalescing testable in the existing node-environment unit suite, with no DOM and no new test dependency.

`App.tsx` reads the width through a ref instead of taking it as an effect/callback dependency, so only the settled width matters.

The rAF shape matches what `useTerminal.ts` already does at `:860` and `:980`.

## Behaviour

Unchanged from the user's side — the width tracks the cursor, and dragging below 80px still auto-collapses. The difference is one update per frame instead of one per event.

## Verification

- 7 unit tests drive the coalescing with a hand-cranked scheduler: burst-collapses-to-one, next frame emits again, `stop()` flushes the pending position, `stop()` cancels the frame, no emit when the pointer never moved, no double-emit.
- `npm test` → 550 passed. The 3 failures (`pty-manager`, `shell-context-menu`) are present identically on unmodified `master` — both need native Windows.

### DIFF
--- a/src/renderer/App.tsx
+++ b/src/renderer/App.tsx
@@ -430,6 +430,12 @@ export default function App() {
     return () => document.removeEventListener('keydown', handleKeyDown);
   }, [shortcuts, commandPaletteOpen]);
   const [sidebarWidth, setSidebarWidth] = useState(DEFAULT_SIDEBAR_WIDTH);
+  // Session writes read the width through this ref, not the state value.
+  // As a dependency, `sidebarWidth` re-subscribed the auto-save listener and
+  // rebuilt the save callback on every intermediate value of a drag; the ref
+  // keeps those stable, so only the settled width matters (issue 07).
+  const sidebarWidthRef = useRef(sidebarWidth);
+  useEffect(() => { sidebarWidthRef.current = sidebarWidth; }, [sidebarWidth]);
 
   // Open tutorial on first launch, unless the welcome screen is disabled in
   // Settings (issue #22). The "seen" flag still prevents re-showing it.
@@ -688,7 +694,7 @@ export default function App() {
         version: 1,
         windows: [{
           bounds: { x: 0, y: 0, width: 0, height: 0 }, // main process fills real bounds
-          sidebarWidth,
+          sidebarWidth: sidebarWidthRef.current,
           activeWorkspaceId: state.activeWorkspaceId,
           workspaces: state.workspaces.map(ws => ({
             id: ws.id,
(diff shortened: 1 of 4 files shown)


amirlehmam:
Merged in 42f9655, shipping in v0.40.0. Thanks — and the framing in the description ("this is churn, not correctness") is why I merged it as-is rather than asking for changes.

The cost chain you traced is the real argument: one `mousemove` → `setSidebarWidth` → workspace relayout → **every** terminal's `ResizeObserver` → `fit()` → `pty.resize()` → ConPTY re-emits its visible screen. At ~60 Hz across every open terminal, a one-second drag was doing far more work in the PTY layer than in the layout layer, and none of it was visible as a bug — which is exactly why it survived this long.

Injecting the frame scheduler instead of calling `requestAnimationFrame` directly is the decision I would have argued for, and you got there first. It is what makes the coalescing testable in the existing node-environment suite with no DOM and no new dependency, and the hand-cranked `fakeFrames` reads better than a mocked global would. `stop()` flushing the pending position is the detail that would have been missed: without it the width settles one frame behind the cursor, which is a visible regression traded for an invisible win.

The `sidebarWidthRef` change carries its own weight too — `sidebarWidth` as a dependency was tearing down and re-registering the auto-save listener on every intermediate value of a drag.

One note for the record: this collided with #136 on `handleSaveSession`. I resolved it in favour of both — the notification string is translated, and the dependency array stays free of `sidebarWidth` since it is read through the ref.

PR #135, opened 2026-07-31
fix(diff): keep the app responsive when a diff pane watches a non-git cwd

### LINKED ISSUE #133: Diff pane on a non-git cwd freezes the app: 2s poll runs a ~3s synchronous full-tree re-read on the main process
## Summary

Opening a diff pane in a workspace whose `cwd` is **not a git repo** pins the main process at ~100% of a core indefinitely and makes the whole app unresponsive. The window stops repainting, Windows marks it "Not Responding", and the named pipe stops answering, so every `wmux <cmd>` call fails with `Error: timeout` — including `identify` and `config show`. Only V1 `ping` keeps working, which makes it look like a pipe/auth bug rather than a stalled event loop.

A home directory is never a git repo, and `~` is the default `cwd` for a new workspace, so this is easy to hit by accident. My saved session had a diff surface as the active tab in both workspaces; every launch froze within seconds.

Reproduced on **v0.39.1** (Windows 11, i5-12450HX, 16 GB). Code below is current `master` (`fbbd375`), unchanged.

## Repro

1. Open a workspace with `cwd` set to a non-git directory with a reasonably large tree (a home directory is ideal).
2. Open a diff pane in it.
3. Within a few seconds the UI stops responding; `wmux identify` starts returning `Error: timeout`.

## Root cause

`DiffPane` polls every 2 seconds:

https://github.com/amirlehmam/wmux/blob/fbbd375b14fb39c263704c1e3455bdb0d71f3e0c/src/renderer/components/Diff/DiffPane.tsx#L124-L143

```ts
// Poll git status every 2 seconds (~50ms per git status call)
...
const id = setInterval(poll, 2000);
```

The `~50ms` budget in that comment holds for the git path. It does not hold for the non-git fallback. When `isGitRepo(cwd)` is false, `getChangedFiles` falls through to `getSnapshotChangedFiles`, which on **every poll**:

1. re-reads **every file** in the snapshot via `readCurrentFile` — `fs.statSync` + `fs.readFileSync`, full contents; and
2. runs a **fresh `walkDir`** over the whole tree to detect new files.

https://github.com/amirlehmam/wmux/blob/fbbd375b14fb39c263704c1e3455bdb0d71f3e0c/src/main/diff-provider.ts#L183-L228

Three things compound:

**1. It is synchronous, on the main process.** `walkDir`/`readCurrentFile` use `readdirSync`/`statSync`/`readFileSync` inside an `ipcMain.handle`. `getChangedFiles` being `async` doesn't help — the sync calls block the event loop, which is why IPC, painting, and the pipe server all stall together.

**2. Each pass costs longer than the poll interval, so the polls overlap and never drain.** Measured by replicating `walkDir` + the snapshot re-read against my home directory:

| | |
|---|---|
| Directories scanned | 3,578 |
| Files collected | 2,000 (`MAX_SNAPSHOT_FILES` cap) |
| Read from disk | 26.6 MB |
| Walk | 1,551 ms |
| Re-read all files | 1,465 ms |
| **Total per poll** | **~3,016 ms** |
| **Poll interval** | **2,000 ms** |

**3. `SNAPSHOT_IGNORE` doesn't exclude platform junk.** It covers `node_modules`, `.git`, `dist`, `build`, `.next`, `__pycache__`, `.venv`, `venv`, plus dotfile dirs. On Windows that misses `AppData`, so a home-directory snapshot happily descends into `AppData\Local`, `AppData\Roaming`, npm caches and installed apps. The 2,000-file cap bounds the *read*, but the walk still enumerates thousands of directories before hitting it — and which 2,000 files you end up with is essentially arbitrary.

## Profile

Attached the debugger to the main process (`wmux.exe --inspect=9230`) and took a CPU profile via `Profiler.start`/`stop`. 9,535 samples over 10 s, self-time:

```
33.1%  readdir  < readdirSync < walkDir
25.8%  stat     < statSync    < readCurrentFile
19.8%  open     < openSync    < readFileSync
 7.2%  readdir  < readdirSync < walkDir
 2.9%  readdir  < readdirSync < walkDir
 2.7%  read     < readSync    < readFileSync
```

~92% of main-process CPU is filesystem syscalls under `walkDir` / `readCurrentFile`. Nothing else is close.

## Confirmation

Removing the two `"type": "diff"` surfaces from `session.json` and restarting, with nothing else changed:

| | Before | After |
|---|---|---|
| Main-process CPU, idle | ~100% of a core, sustained | 0.42% avg over 60 s (0% in the last three 10 s samples) |
| `Responding` | False | True |
| `system.identify` round-trip | timeout (>5000 ms) | 0.7 ms median |
| `pane.list` (round-trips the UI thread) | timeout | 1.3 ms median |
| Main-process RSS | 188 MB, climbing | 111 MB, flat |

## Suggested fixes

Roughly in order of value:

1. **Never let a poll overlap itself.** Schedule the next run only after the previous one resolves (`setTimeout` chain instead of `setInterval`), and back off when a pass exceeds the interval. This alone converts a hard freeze into merely-slow, and is worth doing regardless of the rest.
2. **Make the snapshot path async** (`fs.promises`) so a slow scan degrades into slow diffs rather than a frozen window and a dead pipe.
3. **Stat before reading.** Compare `mtimeMs`/`size` against the snapshot and only `readFile` when they differ. In the steady state nothing has changed, so this removes nearly all 26 MB of re-reads.
4. **Don't re-walk every poll** just to find new files, or do it on a much slower cadence than the content check.
5. **Widen the ignore list** with platform dirs (`AppData`, `Library`, `AppData\Local\Temp`, `$RECYCLE.BIN`, `OneDrive`) and consider a directory-count budget alongside the file cap.
6. **Consider not offering the snapshot fallback at all for very large / home-like trees** — an empty diff pane with "not a git repo" is friendlier than a frozen app.

Happy to open a PR for (1)–(3) if that's useful.

---
*Filed after debugging this on my own machine; all numbers above are measured, not estimated.*

capad-xyz:
Fixes #133 — the three mechanical fixes proposed there: a diff pane on a non-git `cwd` (e.g. a home directory) pinned the main process at 100% of a core, froze the window, and stopped the pipe server answering.

## What changed

**1. The poll can no longer overlap itself** (`DiffPane.tsx`)
`setInterval(poll, 2000)` fired on a fixed clock regardless of how long the previous pass took — with a ~3s snapshot scan against a 2s interval, work queued faster than it drained, permanently. The poll is now a self-scheduling `setTimeout` chain: the next pass is scheduled only after the previous one (files + diff) fully resolves, with the delay stretched to `max(2000, elapsed)` so a slow scan naturally backs the cadence off instead of piling up.

**2. The snapshot path is fully async** (`diff-provider.ts`)
`walkDir` / `takeSnapshot` / `readCurrentFile` used `readdirSync` / `statSync` / `readFileSync` inside `ipcMain.handle`, so the whole scan executed as one uninterruptible block on the main process — which is why IPC, painting, and the named pipe all stalled together. All of it now uses `fs.promises`; a slow scan degrades into slow diffs instead of a frozen app.

**3. Stat before read**
Snapshot entries now store `{ content, mtimeMs, size }` instead of bare content. On each poll a file is `stat`-ed first and its content only re-read when mtime/size actually moved — in the steady state (nothing changed) this eliminates every read; on the tree from #133 that's ~26 MB of `readFileSync` per poll gone. When only metadata moved (`touch`) the stored stat is refreshed so the next poll returns to the cheap path.

Also added, in the same spirit: concurrent scans of the same directory now coalesce onto one in-flight promise (`scansInFlight`), so two diff panes on the same cwd can't double the work.

Not included (deliberately — they're design decisions, per the issue): slower re-walk cadence for new-file detection, widening `SNAPSHOT_IGNORE` with platform dirs, and gating the fallback off for home-like trees.

## Behavior preserved

- Git repos: untouched — same `git status --porcelain` path.
- Baseline semantics: first call snapshots and reports nothing; changes are always relative to the baseline until `resetSnapshot`.
- A file that grows past the 500 KB cap or turns binary is still reported as deleted, as before.
- Depth/file caps, extension filter, ignore list: unchanged.

## Tests

New `tests/unit/diff-provider.test.ts` (the snapshot system previously had no coverage):
- baseline call reports no changes
- modified / added / deleted detection
- **mtime moved but content unchanged → no report, and the skip path holds on the following poll** (the stat-before-read behavior)
- unified diff rendering for a modified file
- concurrent scans coalesce to one result

`npm run typecheck` (both tsconfigs), `npm run lint`, and the full `npm test` suite pass.

## Measured effect

Same machine and tree as #133 (3,578 dirs / 2,000 snapshot files / 26.6 MB):

| | before | after |
|---|---|---|
| Main-process CPU with a diff pane on `~` | ~100% of a core, sustained | idle between polls |
| Window `Responding` | False within seconds | True |
| `system.identify` over the pipe | timeout | sub-ms |

### DIFF
--- a/src/main/diff-provider.ts
+++ b/src/main/diff-provider.ts
@@ -40,7 +40,16 @@ async function isGitRepo(cwd: string): Promise<boolean> {
 // When there's no git repo, we take a snapshot of file contents on first load
 // and diff against it on subsequent polls.
 
-const snapshots = new Map<string, Map<string, string>>(); // cwd → (relPath → content)
+interface SnapshotEntry {
+  content: string;
+  mtimeMs: number;
+  size: number;
+}
+
+const snapshots = new Map<string, Map<string, SnapshotEntry>>(); // cwd → (relPath → entry)
+// One scan per directory at a time: a slow scan coalesces concurrent callers
+// instead of piling up (a home-dir scan can outlast the renderer's poll interval).
+const scansInFlight = new Map<string, Promise<ChangedFile[]>>();
 const SNAPSHOT_EXTENSIONS = new Set([
   '.ts', '.tsx', '.js', '.jsx', '.json', '.py', '.rb', '.go', '.rs',
   '.java', '.c', '.cpp', '.h', '.hpp', '.cs', '.swift', '.kt',
@@ -58,15 +67,15 @@ function shouldSnapshotFile(filePath: string): boolean {
   return SNAPSHOT_EXTENSIONS.has(ext);
 }
 
-function walkDir(dir: string, base: string, files: string[], depth = 0): void {
+async function walkDir(dir: string, base: string, files: string[], depth = 0): Promise<void> {
   if (depth > 5 || files.length >= MAX_SNAPSHOT_FILES) return;
   let entries: fs.Dirent[];
-  try { entries = fs.readdirSync(dir, { withFileTypes: true }); } catch { return; }
+  try { entries = await fs.promises.readdir(dir, { withFileTypes: true }); } catch { return; }
   for (const entry of entries) {
     if (files.length >= MAX_SNAPSHOT_FILES) return;
     if (entry.isDirectory()) {
       if (!SNAPSHOT_IGNORE.has(entry.name) && !entry.name.startsWith('.')) {
-        walkDir(path.join(dir, entry.name), base, files, depth + 1);
+        await walkDir(path.join(dir, entry.name), base, files, depth + 1);
       }
     } else if (entry.isFile() && shouldSnapshotFile(entry.name)) {
       const rel = path.relative(base, path.join(dir, entry.name)).replace(/\\/g, '/');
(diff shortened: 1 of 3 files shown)


amirlehmam:
Merged in ea7c3bb, shipping in v0.40.0. Thank you — for the PR, and for #133, which is one of the best bug reports this project has had.

You did not just say "the app freezes". You attached a debugger to the main process, took a CPU profile, and came back with 92% of self-time in `readdirSync`/`statSync`/`readFileSync` under `walkDir`/`readCurrentFile`, plus a before/after table showing `system.identify` going from timeout to 0.7 ms. That is the difference between a report I have to reproduce and a report I can act on, and it is also what made the symptom legible: only V1 `ping` kept answering, so it *looked* like a pipe or auth bug when it was a stalled event loop. I would have spent a day in `pipe-server.ts`.

The three fixes are the right three, in the right order.

**Never overlapping** is the one that converts a hard freeze into merely-slow, and it is correct regardless of how fast the scan gets. `setInterval` fires on a wall clock that knows nothing about whether the previous pass finished; with a ~3s pass against a 2s interval, work queues faster than it drains and never recovers. Stretching to `max(2000, elapsed)` means the cadence degrades with the tree instead of collapsing.

**Async** is what stops a slow scan from taking IPC, painting and the named pipe down with it. `getChangedFiles` being `async` never helped — the sync calls inside it still owned the thread, which is a distinction worth having in the comments.

**Stat before read** is the one that makes the steady state nearly free. Nothing has changed on most polls, so comparing `mtimeMs`/`size` removes essentially all 26 MB of re-reads. And refreshing the stored stat when only metadata moved is the detail I would expect to be missing — without it a `touch` leaves the file permanently on the expensive path.

`scansInFlight` coalescing two diff panes on the same cwd onto one promise is a good addition that was not asked for.

I also appreciate that you drew a line and said so. Slowing the re-walk, widening `SNAPSHOT_IGNORE` with `AppData`/`OneDrive`/`$RECYCLE.BIN`, and gating the fallback off for home-like trees are design calls, not mechanical fixes, and leaving them out of a fix-the-freeze PR was right. I have kept #133 open to track them — the `AppData` point in particular is real: the 2,000-file cap bounds the read but the walk still enumerates thousands of directories first, and which 2,000 files you end up with is arbitrary.

One observation, not a blocker: `resolveWithin` uses a `startsWith` prefix check, which would accept `C:\foobar` as inside `C:\foo`. That behaviour came over unchanged from `readCurrentFile`, and the paths fed to it come from a walk of `cwd` itself, so it is not reachable today — noting it so it is written down somewhere.

Issue #132, opened 2026-07-30
wmux modifies global claude settings without permission and/or warning


Modestas:
## Problem

`wmux` modifies global `~/.claude` configuration without clearly asking first. If those changes are removed manually, they are added again on the next run.

## Suggestions

1. Move `wmux browser` and `wmux markdown` into optional Claude skills that users can install when needed.
2. Make notification hooks opt-in:
   * Ask on first run before modifying any config.
   * Let users decline permanently.
   * Allow enabling them later from settings.
   * Support both global and per-project installation.

arnonuem:
How can this be reproduced?

Modestas:
Global CLAUDE.md modified by `wmux`: `C:\Users\<YOUR_USER>\.claude\CLAUDE.md`

`https://github.com/amirlehmam/wmux/blob/master/src/main/claude-context.ts`

my quick-fix in fork:
https://github.com/Modestas/wmux/commit/1785ad34a2ec4e4bcf50100045f6ab8c4b51cbe7

amirlehmam:
You were right, and the fix is shipping in **v0.40.0**.

## What was actually wrong

I want to be precise about this rather than just say "added a prompt", because the bug was structural, not an oversight in one function.

`ensureClaudeContext()` and its three siblings were written to be *idempotent* — safe to run on every launch, converging on the same result no matter the starting state. That is a good property, and it is exactly the property that made your deletions futile. Idempotence and consent pull in opposite directions: one says "always reach the same end state", the other says "the user's decision is part of the state". The code implemented the first and had no representation for the second at all. There was nowhere for a "no" to live, so there was nothing for the next launch to read.

Your fork's quick-fix went straight at that, and the shape you chose is the shape I built.

Also worth stating plainly: it was four separate writes, not one, and the issue title undersells it. Per launch, wmux was touching `~/.claude/CLAUDE.md`, `~/.claude/settings.json` (twice — hooks *and* an MCP server entry, plus force-disabling the official `chrome-devtools-mcp` plugin), `~/.claude/plugins/`, and — undocumented anywhere — `~/.config/opencode/AGENTS.md` and `~/.config/opencode/plugin/`. If you had only removed the Claude files you would still have been carrying two more.

## What shipped

A decision stored in **wmux's own** settings file (`%APPDATA%\wmux\settings.json`), never in the files being written:

```
unset     nothing written yet — ask
granted   apply the features that are on
declined  write nothing, and take back anything an earlier version wrote
```

[...]

PR #136, opened 2026-07-31
feat(i18n): full renderer i18n audit — translate every remaining hardcoded string (es/fr/it/zh)

simplyBarbe:
## Summary
Completes the i18n coverage of the renderer, translating every user-facing
string that was still hardcoded in English into Spanish, French, Italian,
and Chinese. Also fixes a bug where main-process file errors always won
over the renderer's translated fallback.

- Audited `src/renderer/` end-to-end for hardcoded text not routed through
  `useT()`/`translate()`, area by area:
  SplitPane, Tutorial, Sidebar, Settings sub-panels, ErrorBoundary,
  OrchestrationPanel, ShortcutCheatSheet, Diff pane, Terminal
  (CopyMode/FindBar/bell), Titlebar notifications, ConfirmCloseDialog,
  Browser AddressBar, CommandPalette/WorkspaceContextMenu/GeneralSettings
  stray misses, `App.tsx` (OS notifications, default session titles,
  sidebar/browser tooltips, broadcast-input banner), default
  workspace/session titles, the sidebar "+N more" agent summary,
  notification sound names, and a stray SplitPane tab-menu badge.
- For non-hook contexts (pure functions, Zustand store slices, class
  components) threaded an optional `t` parameter with an `identityT`
  default, so existing callers/tests keep working untouched.
- **Bug fix:** `markdown-file.ts` read/write errors always carried a
  truthy English `error` string, so the renderer's
  `t('markdown.error.*', ...)` fallback never fired. Added a stable
  `code` field alongside it — CLI/pipe callers still see the English
  message verbatim, the renderer now maps `code` to a translated message.
- Deliberately left untranslated (consistent with existing precedent):
  proper nouns — search engine names, shell names, placeholder terminal
  theme names — and main-process-only surfaces (native dialogs, tray-less
  update prompt, Explorer context-menu verb) that would need a separate
  language-sync mechanism between main and renderer.

## Test plan
- [x] `npx tsc --noEmit -p .` — clean
- [x] `npm test` — 567/567 passing (54 files)
- [x] `npm run build:main` — compiles clean
- [x] `npm run lint` — no new errors/warnings introduced (verified via
      `git diff` that all 19 errors / 13 warnings pre-date this branch)
- [x] Manually switched the language selector across all 5 locales in
      each translated area to confirm fallback-to-English works for any
      partial coverage


amirlehmam:
Merged in 0f99679, shipping in v0.40.0. Thank you — 3,110 lines across 49 files is a lot of unglamorous work, and it closes the gap that made wmux's i18n feel half-done: the language selector existed, but switching it left English scattered through Settings, the Tutorial, CopyMode, the Diff pane and the notification panel.

Three things I want to single out.

**The `markdown-file.ts` bug is the real find here.** Read/write errors always carried a truthy English `error` string, so the renderer's `t('markdown.error.*')` fallback could never fire — the translation existed and was structurally unreachable. Adding a stable `code` alongside the message, rather than replacing it, is the right shape: CLI and pipe callers still get the English text verbatim, and the renderer maps the code. Swapping the message for a code would have broken every non-UI consumer to fix a UI bug.

**Threading an optional `t` with an `identityT` default** through pure functions, store slices and class components is the part most people get wrong. The alternative — reaching for a hook where hooks do not belong, or a module-level singleton — would have made those functions untestable and order-dependent. Existing callers and tests kept working untouched, which is the proof it was done right.

**Knowing what not to translate.** Search-engine names, shell names, placeholder theme names, and main-process-only surfaces that would need a language-sync mechanism between main and renderer — leaving those and saying so explicitly is better judgement than blanket coverage. A translated `PowerShell` helps nobody.

I resolved one conflict against #131 (frame-coalesced sidebar drag), which landed on `handleSaveSession` first: the notification takes your translated string, and the dependency array stays free of `sidebarWidth` since it is now read through a ref. `npx tsc --noEmit` clean on both configs, 624/624 tests passing on native Windows.

I have added five keys on top of yours in this release, in all five locales — `settings.integration.*`, for the new agent-integration consent panel (#132).

PR #138, opened 2026-07-31
fix(diff): bound the snapshot walk and slow the new-file discovery cadence

capad-xyz:
Follow-ups **(1)** and **(2)** from #133, which were deliberately left out of #135 as design calls rather than bug fixes. **(3)** — whether to offer the fallback at all for home-like trees — is still yours; happy to implement whatever you land on.

## (1) Bound the walk, not just the read

`SNAPSHOT_IGNORE` gains the platform directories: `AppData`, `Library`, `Application Data`, `$RECYCLE.BIN`, `System Volume Information`, `OneDrive`, `Recovery`.

More importantly, `MAX_SNAPSHOT_DIRS` (500) is now threaded through `walkDir` as a mutable budget shared across the recursion, so traversal is bounded independently of the file cap. This is the half you identified as the important one: the 2,000-file cap bounds how much we *read*, but ~3,578 directories were still enumerated before it bit, and the resulting 2,000 files were an accident of traversal order.

## (2) Slower discovery cadence

The re-walk that finds *new* files now runs every 5th poll (`REWALK_EVERY_N_POLLS`). The stat-based check for *existing* files still runs every poll.

**Behaviour change worth calling out explicitly:** a newly added file now appears within ~10s instead of ~2s. Modifications and deletions are unchanged — they come from the snapshot map and are still detected on the very next poll. If you want a different number, it's a one-line constant.

There's a trap in this change that isn't obvious: because newly discovered files are not in the snapshot map, running discovery only every Nth poll makes them *flicker* — present on poll 5, absent on 6–9, present on 10. So discovered additions are now remembered per-cwd in `addedPaths` and re-reported on every poll (and dropped when the file goes away). `resetSnapshot` clears that alongside the snapshot. Two tests pin this: one that an added file keeps appearing across 12 consecutive polls, one that it disappears once deleted.

## Also: the prefix check you noted

`resolveWithin` used `resolved.startsWith(path.resolve(cwd))`, which accepts `C:\foobar` as inside `C:\foo`. The untracked-file branch of `getFileDiff` had the identical test three lines away. Both now go through one `isWithin` helper built on `path.relative`.

Neither was reachable — as you said, the paths come from a walk of `cwd` itself — so this is hygiene, not a security fix. Flagging it as slightly outside what you asked for; it seemed odd to fix one and leave its twin adjacent to it.

## Tests

`tests/unit/diff-provider.test.ts` grows from 5 to 9 cases:
- added files surface once the periodic re-walk runs
- an added file keeps being reported between re-walks (no flicker)
- a discovered file drops out of the list once deleted
- `AppData/**` is never reported, across several polls

The pre-existing "modified, added, deleted" case was split in two, because added-file detection is deliberately no longer immediate and the old case asserted that it was. Modified/deleted still assert next-poll detection.

Full suite: 61 files, 667 tests passing. `npm run typecheck` and `npm run lint` clean.

### DIFF
--- a/src/main/diff-provider.ts
+++ b/src/main/diff-provider.ts
@@ -58,24 +58,49 @@ const SNAPSHOT_EXTENSIONS = new Set([
   '.sh', '.bash', '.zsh', '.ps1', '.bat', '.cmd',
   '.sql', '.graphql', '.proto', '.xml', '.svg',
 ]);
-const SNAPSHOT_IGNORE = new Set(['node_modules', '.git', 'dist', 'build', '.next', '__pycache__', '.venv', 'venv']);
+const SNAPSHOT_IGNORE = new Set([
+  // Build/dependency output
+  'node_modules', '.git', 'dist', 'build', '.next', '__pycache__', '.venv', 'venv',
+  // Platform directories. A home directory is never a git repo, so it lands on
+  // the snapshot path; without these the walk descends into caches and installed
+  // applications, burning the directory budget on files nobody is editing.
+  'AppData', 'Library', 'Application Data', '$RECYCLE.BIN', 'System Volume Information',
+  'OneDrive', 'Recovery', 'ntuser.dat',
+]);
 const MAX_SNAPSHOT_FILE = 500_000; // 500KB per file
 const MAX_SNAPSHOT_FILES = 2000;
+// The file cap bounds how much we READ, not how much we WALK: on a home
+// directory ~3,500 directories get enumerated before the file cap bites, and
+// which files you end up with is then an accident of traversal order. Budget
+// directories too, so the walk itself is bounded.
+const MAX_SNAPSHOT_DIRS = 500;
+// Re-walking to discover new files is far more expensive than stat-ing the
+// files already known, and new files appear far less often than they change.
+// Run discovery on a multiple of the content check instead of every poll.
+const REWALK_EVERY_N_POLLS = 5;
+const pollCounts = new Map<string, number>(); // cwd → polls since last re-walk
+const addedPaths = new Map<string, Set<string>>(); // cwd → files discovered since the baseline
 
 function shouldSnapshotFile(filePath: string): boolean {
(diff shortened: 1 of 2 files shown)


amirlehmam:
Merged in 970f7b8. Verified locally first: 61 files / 667 tests, typecheck clean — exactly what you reported.

Both halves are right, and (1) is the one I'd have got wrong on my own. #135 bounded how much we *read*; it left the walk itself unbounded, so the 2,000 files we ended up with were an accident of traversal order — a `MAX_SNAPSHOT_FILES` that bites after 3,578 directories isn't a budget, it's a coincidence. Threading `MAX_SNAPSHOT_DIRS` through the recursion as shared mutable state is the correct shape for that, and it's the kind of thing that reads as a small diff and isn't.

On (2): the flicker trap is the part of this PR I'm most glad you wrote down. Making discovery periodic while modification detection stayed per-poll creates two classes of file with different visibility rules, and the failure — present on poll 5, gone on 6–9, back on 10 — would have looked like a rendering bug for a long time before anyone suspected the poll cadence. `addedPaths` plus the two tests that pin it across 12 consecutive polls is the right answer; splitting the old "modified, added, deleted" case in two rather than loosening it was also right, since the whole point is that those three no longer share a timing contract.

~10s to notice a new file is fine. Modifications and deletions are what people actually watch a diff pane for, and those are unchanged at one poll.

The `resolveWithin` prefix check: taking it. You flagged it as slightly outside scope and it is, but leaving one of an adjacent pair fixed is worse than fixing both, and `path.relative` is the version that doesn't need a comment explaining why `C:\foobar` isn't inside `C:\foo`.

On (3) — I'm not going to offer the fallback for home-like trees. A diff pane pointed at `%USERPROFILE%` has no useful answer to give; the honest thing is to say so rather than to compute a bounded-but-arbitrary subset that looks like a real answer. I'll open that separately if it turns out people actually do it by accident.

Thank you — that's #133, #135 and now #138, and the standard hasn't dropped once.

Issue #121, opened 2026-07-27
browser.click/get_text always fail with ref_not_found right after a fresh browser.snapshot


yatran911:
## Summary

`browser.click` and `browser.get_text` always fail with `ref_not_found`, even immediately after a fresh `browser.snapshot` returns a valid ref tree with a nonzero `refCount`. Confirmed this is not related to the CLI, PowerShell, connection reuse, or timing - reproduced by talking to the named pipe directly.

## Repro steps (raw JSON-RPC over the named pipe, bypassing the CLI entirely)

Using a single persistent TCP connection to `\\.\pipe\wmux` (so it's not a "new connection per request" issue):

1. Send `{"method":"browser.snapshot","params":{"caller":"<browser-surface-id>"},"id":1,"token":"<pipe-token>"}`
   -> returns `{"tree":"@e1: RootWebArea ...", "refCount": 300}` (or similar, tree/refCount vary by page).
2. Immediately (same connection, no delay) send `{"method":"browser.click","params":{"ref":"e2","caller":"<same-surface-id>"},"id":2,"token":"<pipe-token>"}`
   -> returns `{"error":{"code":-32000,"message":"ref_not_found"}}`

Also tried via the CLI (`wmux browser snapshot` then `wmux browser click e2`) - same result.

## What I ruled out

- **Not a connection-lifetime issue**: sent snapshot+click over the SAME socket without closing it between requests - still `ref_not_found`.
- **Not a timing issue**: sent click with zero artificial delay right after snapshot - still fails. Also tried inside one long-lived Node process reusing the connection.
- **Not a caller/surface mismatch**: explicitly passed the same `caller` (browser surface id, matching `WMUX_SURFACE_ID` / the id returned by `list-surfaces` for the `type: "browser"` entry) to both calls.
- **Not a ref format issue**: tried `"e2"`, `"2"`, `2`, and `"@e2"` as the `ref` value. All bare formats fail with `ref_not_found`; only the `@`-prefixed form fails differently, with `Invalid parameters` (consistent with docs saying refs take no `@` prefix).
- `browser.get_text` with the same ref also fails with `ref_not_found`, so this isn't click-specific - it looks like every ref-consuming command can't resolve refs that `browser.snapshot` just generated.

## Environment

- wmux (Windows), CLI at `resources/cli/wmux.js`
[...]

Issue #62, opened 2026-06-26
Concurrent agents share a single browser window, causing interference


AxelM-MSK:
Problem

  When multiple agents are active at the same time, they all control the same
  browser window. There's no isolation between sessions, so one agent's actions
  disrupt another's in-progress task.

  Current behavior

  - All agents target the one shared browser window.
  - When Agent B issues a command (navigate, click, type), it acts on the window   
  Agent A is mid-task in.
  - Agent A's state is clobbered — page navigated away, focus lost, refs
  invalidated — and its task fails or produces wrong results.

  Expected behavior

  Each agent should operate against its own isolated browser context so concurrent 
  work doesn't collide. Also the browser windows should spawn in the same workspace as the agent running it

  Impact

  - Agents silently corrupt each other's work; failures are hard to trace back to  
  the real cause (a different agent moved the page).
  - Makes any parallel/multi-agent browser workflow unreliable.

amirlehmam:
Fixed in **v0.12.0** 🎉 — thanks for the clear write-up, this was exactly the right root cause.

The problem was a single shared CDP connection: whichever browser pane attached last "owned" the Chrome DevTools debugger, so a second agent's `wmux browser` commands would steal the connection and act on the first agent's page.

**What changed:**
- The CDP bridge is now **multi-target** — every browser pane keeps its own independent debugger attachment and accessibility ref map. No more stealing.
- Each caller (an agent's terminal, via `WMUX_SURFACE_ID`, which the CLI now sends with every `browser.*` command) is **bound to its own browser surface**, created **in that caller's workspace** — addressing both asks in your issue. A second agent that finds the first's browser already owned spins up its own instead of colliding.
- Manual (human) `wmux browser` use is unchanged — with no agent caller it falls back to the shared browser as before.

So two agents running browser tasks concurrently now each drive a separate, isolated browser pane in their own workspace. Available in the v0.12.0 zip on the releases page (right-click → Unblock after download). Let me know how it holds up in your multi-agent runs!

Issue #149, opened 2026-08-07
bug: opencode-plugin folder not included in install zip


kudlosh:
The _resources/opencode-plugin/_ folder is not present in the distributed .zip release.

### Steps to reproduce:
1. Download the latest wmux-<VERSION>-win-x64.zip from the releases page
2. Extract it
3. Look in resources/ — there is no opencode-plugin/ folder

**Expected:** resources/opencode-plugin/wmux.js should be present in the release zip, the same way wmux-orchestrator/ is.

**Why it matters:** the OpenCode plugin is required for OpenCode integration, and is missing from the zip.

amirlehmam:
Confirmed and fixed in **v0.47.0**. Your reproduction was exact — the folder was never in the zip, in any release since the plugin was introduced.

## Cause

`resources/opencode-plugin` was simply absent from `extraResources` in `electron-builder.json`, so it was never copied next to `app.asar`. One line:

```json
{ "from": "resources/opencode-plugin", "to": "opencode-plugin" }
```

What made it invisible is the fallback in `ensureOpencodePlugin()`. It resolves `process.resourcesPath/opencode-plugin/wmux.js` when packaged and a repo-relative path otherwise, and on a miss it warns to a console no user sees and returns. So it worked perfectly in `npm run dev` and silently did nothing in every install — the failure mode had no symptom beyond the one you noticed by looking.

## Why the test suite didn't catch it

There was already a `packaging.test.ts`. It was written after #81 (`wmux-hook.js` missing from `extraResources`, which broke every Claude Code hook) and it asserted two literal entries — the two that had already broken. A test that enumerates the failures it knows about can only ever catch those, and this walked straight past it.

That's now derived instead of enumerated: the test scans `src/main/` for `process.resourcesPath` references and requires each one to be listed in `extraResources`. Adding a runtime resource without packaging it fails `npm test`, which gates the release build, so a tag that would ship this bug can't become a download. Verified it fails without the fix before trusting it.

## Verified in the artifact, not just the config

```
resources/opencode-plugin/wmux.js  (2366 bytes)
```

read out of the published `wmux-0.47.0-win-x64.zip`.

## Upgrading

[...]

PR #153, opened 2026-08-12
fix(cli): outwait the main process on browser commands, and say what timed out

capad-xyz:
## Summary

`wmux browser open <url>` fails with a bare `Error: timeout` after ~5s on any page that takes longer than that to load — while the navigation itself goes on to succeed. The CLI applies one flat 5s deadline to every V2 method, but the main process is allowed to spend considerably longer serving a browser command.

The deeper problem is the second-order one: because the client deadline is *shorter* than the server's own budget, the server's real error message can never reach the user. A bare `timeout` is the only thing a slow or broken browser command can print, and it reads like a broken install.

## Repro

Against a healthy wmux (v0.46.0, Windows 11), with a browser pane open and CDP attached. Any page slower than 5s will do; a local server makes it deterministic:

```js
// slow-server.js — headers immediately, body finishes at 12s
require('http').createServer((req, res) => {
  res.writeHead(200, { 'Content-Type': 'text/html' });
  res.write('<html><body>loading');
  setTimeout(() => res.end('done</body></html>'), 12000);
}).listen(8931, '127.0.0.1');
```

```
$ wmux browser open http://127.0.0.1:8931/
Error: timeout
exit=1  elapsed_ms=5157
```

The navigation was not the problem. Sending the identical request over the pipe by hand, with a 40s deadline instead of 5s:

```
replied after 12017ms: {"result":{"ok":true},"id":1}
```

The server answered correctly. The CLI just stopped listening at 5s.

## Root cause

`sendV2` hard-codes 5000ms and rejects with `new Error('timeout')`:

https://github.com/amirlehmam/wmux/blob/fae21a5bbdf2edcd1c41d4c0b5372aab56caef83/src/cli/wmux.ts#L93

Three main-process budgets sit at or above that number, so the CLI is structurally guaranteed to lose the race:

| Server-side wait | Budget | Source |
|---|---|---|
| `navigate()` waiting for `did-finish-load` | **30000ms** | [cdp-bridge.ts#L212](https://github.com/amirlehmam/wmux/blob/fae21a5bbdf2edcd1c41d4c0b5372aab56caef83/src/main/cdp-bridge.ts#L212) |
| `wait()` polling for a ref | **10000ms** | [cdp-bridge.ts#L373](https://github.com/amirlehmam/wmux/blob/fae21a5bbdf2edcd1c41d4c0b5372aab56caef83/src/main/cdp-bridge.ts#L373) |
| readying a browser (pane split, then poll for CDP attach) | **5000ms** | [v2-browser.ts#L33-L37](https://github.com/amirlehmam/wmux/blob/fae21a5bbdf2edcd1c41d4c0b5372aab56caef83/src/main/v2-browser.ts#L33-L37), [#L100](https://github.com/amirlehmam/wmux/blob/fae21a5bbdf2edcd1c41d4c0b5372aab56caef83/src/main/v2-browser.ts#L100) |

Two consequences, and the second is the one that costs debugging time:

1. A command that succeeds late is reported as a failure. `browser open` is effectively nondeterministic — it passes or fails on page speed, and when it "fails" the navigation has still happened.
2. **The server's own diagnosis is unreachable.** `Could not open browser panel`, `browser_not_open` and the careful `ref_not_found: … the last snapshot of this browser has @e1..@eN` message from #121 all arrive after the CLI has hung up. The user gets `timeout` instead.

### A note on the obvious-looking diagnosis

I first went after this as a missing-caller bug: no `WMUX_SURFACE_ID` outside a pane → no `caller` attached → main can't route the request. That turns out **not** to be what is happening, and I want to flag it since it is the intuitive reading.

With no caller, `resolveBrowserWcId` deliberately falls back to `legacyWcId()` — the documented shared-browser path for manual human use ([v2-browser.ts#L8-L9](https://github.com/amirlehmam/wmux/blob/fae21a5bbdf2edcd1c41d4c0b5372aab56caef83/src/main/v2-browser.ts#L8-L9), [#L68](https://github.com/amirlehmam/wmux/blob/fae21a5bbdf2edcd1c41d4c0b5372aab56caef83/src/main/v2-browser.ts#L68)). That path works fine:

```
$ wmux browser open https://example.com     # no WMUX_SURFACE_ID anywhere
{ "ok": true }                               # 1709ms
```

And the failure reproduces just as reliably *with* a valid caller set:

```
$ WMUX_SURFACE_ID=surf-617012f2-… wmux browser open http://127.0.0.1:8931/
Error: timeout
exit=1  elapsed_ms=5124
```

Same result, 5124ms. The caller is irrelevant to this failure; the deadline is the whole story. Anywhere setting `WMUX_SURFACE_ID` appears to fix it, what actually changed is which resolution path ran and whether it happened to finish inside 5s.

## The fix

**Derive each browser verb's client deadline from the budget the server may spend on it**, so the main process always loses the race and its own error is what surfaces. `BROWSER_READY_MS + verbBudget + slack`: 40s for `open`, 20s for `wait` (or the explicit `ms` plus headroom), 10s for everything else. The 5s default is unchanged for every other V2 method.

**Make the timeout message diagnostic.** It now names the method and the deadline actually waited, and says the command may have completed anyway — because it may have.

**Make the two cdp-bridge timeouts earn their keep.** They also threw a bare `new Error('timeout')`. That was invisible before (the CLI's own timeout always fired first); now that it is what the user reads, it should say which operation stalled, for how long, and on what. This is the same treatment `ref_not_found` got in #121.

**Add `--surface` to the browser verbs**, matching `send` / `read-screen` / `agent-activity` / `report-agent`. A shell wmux did not spawn has no `$WMUX_SURFACE_ID` and so had no way to say which pane's browser it meant; the legacy fallback picks the most-recently-attached one, which is arbitrary when several exist. The flag is stripped before the verb reads its positional args, so `browser type e5 --surface surf-x hello world` still types `hello world`.

## Options I considered

1. **Fail fast when there is no caller, with an actionable message.** Rejected: it would turn a working path into an error. The no-caller fallback is deliberate and, as shown above, works — and it is not what breaks here.
2. **`--surface` alone.** Useful, and included, but it does not fix the reported bug at all: the failure reproduces with a valid caller.
3. **Fall back to the active surface when no caller is set.** Rejected, and I would push back on it: `resolveBrowserWcId` binds a caller to its own browser and records it in `boundBrowserSurfaces` so a second agent never adopts the first agent's browser. Silently substituting "the active surface" for an absent caller would let an agent-less invocation adopt a browser an agent already owns — exactly the #62 clobbering, reintroduced through the back door.
4. **Raise the flat deadline for everything.** Rejected: it would slow the failure path for every method that legitimately answers in milliseconds, and it papers over the client/server mismatch rather than fixing it.

## What I deliberately did not change

- **The isolation contract (#62).** No change to `resolveBrowserWcId`, to how callers bind to browser surfaces, or to the no-caller fallback. `--surface` only supplies from a flag what a pane already supplies from the environment.
- **The server-side budgets themselves.** 30s/10s/5s are the main process's calls to make; the CLI now accommodates them instead of contradicting them. The tests read those numbers out of the main-process source rather than restating them, so changing one there fails here rather than silently reintroducing the mismatch.
- **The 5s default for non-browser methods.**
- **`browser.batch`**, which the CLI does not expose.
- **The diff-provider snapshot code**, untouched.

## Tests

`tests/unit/browser-timeout.test.ts` (new, 12 cases) and 2 added to `tests/unit/cdp-bridge.test.ts`. Following the approach in `packaging.test.ts`, the deadline assertions are *derived* from the main-process source rather than restated, so a raised default in `cdp-bridge.ts` fails the test instead of quietly recreating the bug.

Confirmed the tests fail against the old behaviour: pinning `browserDeadline` back to a flat 5000 fails exactly the 4 cases that describe the bug.

To make the CLI importable by tests at all, `main()` is now guarded by `require.main === module`. Nothing imports `src/cli/wmux.ts` as a module today — it is only ever executed as a script, where the guard is true. I kept everything in the one file rather than extracting a module, since the release process copies `dist/cli/wmux.js` individually and a new sibling would not be packaged.

```
Test Files  1 failed | 72 passed (73)
Tests       1 failed | 848 passed (849)
```

The one failure is `tests/unit/cdp-proxy.test.ts > falls back to a free port instead of raising an uncaught error`. It fails identically on a clean `master` checkout with nothing applied, so it is pre-existing and not from this PR — it looks environment-dependent (the default port being free on this machine). I left it alone rather than fold an unrelated fix in here.

## Verification

Built and run against a live wmux instance, same command, before and after:

```
### BEFORE (shipped 0.46.0 CLI)
$ wmux browser open http://127.0.0.1:8931/
Error: timeout
exit=1  elapsed_ms=5157

### AFTER (this branch)
$ node dist/cli/wmux.js browser open http://127.0.0.1:8931/
{ "ok": true }
exit=0  elapsed_ms=12212
```

Also verified live: `--surface <id>` drives the named pane's browser from a shell wmux did not spawn; `browser snapshot` still returns in ~215ms (no added latency on the fast path); non-browser methods unaffected; unknown verbs still exit 1.

On a URL that never finishes loading, the CLI now waits and receives the **server's** answer at 30s rather than its own at 5s — which is the plumbing fix working end to end. The running instance there was 0.46.0, so the text it returned was still the old bare `timeout`; the new cdp-bridge wording is covered by unit tests rather than that live run, since exercising it would have meant restarting the instance.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -396,6 +396,8 @@ wmux read-screen [--lines N] [--surface <id>] | trigger-flash
 wmux browser open <url> | snapshot | click eN | type eN <text>
 wmux browser fill eN <value> | get-text | screenshot | eval <js>
 wmux browser back | forward | reload
+wmux browser <verb> [--surface <id>]   # whose browser to drive; defaults to
+                                       # $WMUX_SURFACE_ID inside a pane
 
 # Declared agent state (issue #128) — blocked / working / idle, no screen scraping.
 # Surface defaults to $WMUX_SURFACE_ID, so an agent inside a pane needs no id.
--- a/src/cli/wmux.ts
+++ b/src/cli/wmux.ts
@@ -75,7 +75,35 @@ function sendV1(command: string): Promise<string> {
   });
 }
 
-function sendV2(method: string, params: Record<string, any> = {}): Promise<any> {
+/**
+ * How long to wait for a V2 reply before giving up.
+ *
+ * This deadline has to stay LARGER than whatever budget the main process spends
+ * serving the same request. When it is shorter the CLI loses a race it should
+ * never have been in: a command that succeeds late is reported as a failure, and
+ * the server's own diagnosis ('Could not open browser panel', 'browser_not_open',
+ * 'ref_not_found: …') is discarded unread because it arrives after we hung up.
+ * Only the browser verbs currently need more than this — see BROWSER_CMDS.
+ */
+const DEFAULT_V2_TIMEOUT_MS = 5000;
+
+/**
+ * What a stalled request says when it gives up.
+ *
+ * The bare 'timeout' this used to reject with named neither the method nor the
+ * deadline, so an operation that was merely slow was indistinguishable from a
+ * broken install — and since the deadline was also shorter than the server's own
+ * budget, it was usually the *only* thing a slow browser command ever printed.
+ */
+export function timeoutMessage(method: string, timeoutMs: number): string {
+  return `${method} timed out after ${timeoutMs}ms — wmux accepted the request but sent no reply. The command may still have completed.`;
+}
+
+function sendV2(
+  method: string,
+  params: Record<string, any> = {},
+  timeoutMs: number = DEFAULT_V2_TIMEOUT_MS,
+): Promise<any> {
   // Every command carries the caller's surface (WMUX_SURFACE_ID). Browser
   // commands use it to route each agent to its OWN browser pane, so concurrent
   // agents no longer share and clobber one browser window (issue #62); the
@@ -90,7 +118,10 @@ function sendV2(method: string, params: Record<string, any> = {}): Promise<any> 
       client.write(request + '\n');
     });
     let data = '';
-    const timer = setTimeout(() => { client.end(); reject(new Error('timeout')); }, 5000);
+    const timer = setTimeout(() => {
+      client.end();
+      reject(new Error(timeoutMessage(method, timeoutMs)));
+    }, timeoutMs);
     client.on('data', (chunk) => {
       data += chunk.toString();
       if (data.includes('\n')) {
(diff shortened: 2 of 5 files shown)


capad-xyz:
Some independent corroboration for the `--surface` half of this, from a separate heavy-usage session on the same machine that ran into it without knowing the PR existed:

> `browser open <url> --surface <id>` silently ignores the flag and navigates the active surface instead. That's how I knocked your other session off its page. Silent ignore is worse than an error.

and, separately:

> `WMUX_SURFACE_ID` works for browser commands but is undocumented for them. The help only mentions it under `report-agent`. It's the single most useful thing I learned all night and I found it by guessing.

Both are what this PR addresses — the flag, and the `(surface defaults to $WMUX_SURFACE_ID inside a pane)` note now on the `browser` usage line and in the CLI reference. Worth noting the failure mode they describe is the concurrent-browser collision from #62 arriving through a different door: the isolation routing works, but with no way to *address* it from outside a pane, two sessions end up on one browser anyway.

That session surfaced four other things. Three held up on investigation and are filed separately — #154 (PowerShell loses argument quoting through `wmux.cmd`; `>` in a free-form argument silently redirects), #155 (a dead browser webContents leaves a permanently stale CDP target), #156 (`wmux browser` with no subcommand prints `undefined` instead of usage). #156 is in `cmdBrowser`, the same function this PR rewrites — happy to fold it in here instead if you'd prefer that to a separate change.

amirlehmam:
Merged, unchanged, and shipped in **0.50.0**. Thank you — this is the right diagnosis and the right fix.

The part I want to call out is not the deadline arithmetic, it's the second-order framing: *a client deadline shorter than the server's own budget makes the server's error message unreachable*. Three careful diagnostics — `Could not open browser panel`, `browser_not_open`, and the `ref_not_found: … the last snapshot of this browser has @e1..@eN` text from #121 — could never be printed by the CLI, because it had already hung up. Every one of them arrived as `timeout`. That is worth more than the reported symptom, and it is why `browserDeadline` deriving from `BROWSER_READY_MS + verbBudget + slack` is the correct shape rather than "raise 5000 to 40000".

Also right on the two things you *didn't* do:

- Rejecting "fall back to the active surface when no caller is set" is correct, and for exactly the reason you gave: it would let an agent-less invocation adopt a browser an agent already owns, which is #62 coming back through the side door. `--surface` supplies from a flag what a pane supplies from the environment, and nothing else changes.
- Reading the budgets out of the main-process source in the tests instead of restating them. `packaging.test.ts` earned that pattern the hard way (#149 walked straight past a test that enumerated the failures it already knew about); a test that pins a number *by copying it* is a test that agrees with the bug once someone edits the other file.

Verification here before merging: full suite `73 files / 849 tests` green on your branch, including `tests/unit/cdp-proxy.test.ts`. That one is environment-dependent as you suspected — it passes on this machine and failed on yours, which is a real (pre-existing) flake worth its own issue, not something to have folded in here.

On your offer in #156: I took it separately, since it touches `cmdAgent`, `cmdPane` and `cmdLayout` too and I didn't want a usage-message change riding inside a timeout fix. See the comment there.

Issue #151, opened 2026-08-11
Claude status update inconsistencies


derekneely:
First off, love what you've done. I use cmux on my mac and found this repo as I'm forced to use Windows at work.

Thought I'd post something that I've found after installed. The statuses to not seem to line up with with what Claude is doing very often. More times than not the workspace will say Idle when in fact an agent is running some stuff. Please see screenshot. In addition when you first start a Claude session it will say 'Running' even though it is a brand new instance that is doing nothing. Also, after going to a 'Needs you' workspace, answering questions or following up. The window does not come out of that state. It seems to be stuck in a 'Needs you' status even after resuming work. I find myself having to just click through tabs to see where things are instead of being able to rely on the actual status of the workspace.

<img width="1410" height="837" alt="Image" src="https://github.com/user-attachments/assets/2901f385-5d00-4b6a-9918-57f7e0d7fbf2" />

derekneely:
Not sure if interested but I forked the code and working through some updates on the status handling now. Will test through for a work day or two and can provide a PR back to you if that helps you out.

amirlehmam:
Thank you — this is a good report, and all three symptoms turned out to be one root cause plus three smaller bugs sitting on top of it. Fixed in **0.48.0**.

## The root cause: wmux only listened for things ENDING

wmux drives its status from Claude Code hooks rather than from scraping the TUI. It registered four of them:

```
PostToolUse    a tool finished
Notification   Claude wants you
Stop           the turn finished
SubagentStop   a subagent finished
```

Three of those are terminal events. **Nothing ever said work had started.** So the state machine only learned a turn was in flight when the first tool *completed* — and everything before that resolved to `idle`:

- you send a message, Claude thinks for 40 s before touching a tool → **Idle**
- Claude runs one three-minute `npm test` → **Idle** until it finishes
- a turn that reasons and writes prose without calling a tool at all → **Idle**, start to finish

That's your "more times than not it says Idle when an agent is running". It was never a timing or a refresh problem; the event that would have said otherwise was not being listened for.

0.48.0 registers the opening half of the lifecycle:

| hook | what it now tells wmux |
|---|---|
| `SessionStart` | this pane is an agent session, currently idle |
| `UserPromptSubmit` | the human replied — turn started, and any wait is over |
| `PreToolUse` | a tool is starting (matcher-less, so it covers *every* tool, not just the tracked ones) |
| `SessionEnd` | Claude Code exited — release the pane |

[...]

PR #159, opened 2026-08-13
fix(agent-state): status path handling refinement and edge cases (#151)

derekneely:
Follow-up to #151, on top of the 0.48.0 rewrite.

That change is what made the rest of this findable — once wmux could hear a turn *start*, the status became reliable enough that the remaining wrong readings had a shape instead of feeling random. I've been running it as my daily driver since, first on 0.48.0 and then on 0.50.0, mostly with several Claude panes going at once. Four narrower paths turned up in that use, and each one still surfaces as a symptom from my original report.

None of this changes the model 0.48.0 established. It's the same hook-driven design, extended to a few shapes that don't come up until you lean on it: one agent pane instead of several, background shells, parallel subagents, and interrupts.

Four commits, in dependency order. They're independent enough to be reviewed one at a time, and I'm happy to split any of them out.

| # | Edge case | What you see |
|---|---|---|
| 1 | Single-session path doesn't reach `workingSessions` | one agent pane reads **Idle** mid-turn |
| 2 | Three cases in the hook bridge | turns age out of `working`; **Needs you** on a pane nobody touched; Escape leaves `working` set |
| 3 | `SubagentStop` can arrive after `Stop` | pane stays on **Running** after the agent is idle |
| 4 | Tool-lifecycle events clear `awaitingHuman` | a question on screen while the row says **Running…** |

---

## 1. Read declared session state in the single-session case

`workingSessions` — the count of panes whose *declared* state is `working` — is consulted inside the `sessionCount >= 2` branch of `claudeStatusText`. The single-pane case, which is the common shape, doesn't reach it yet.

It falls through to priority 1, the tool label. `PostToolUse` fires when a tool *finishes*, so during a stretch of thinking or one slow tool the label expires while the turn runs on. The chain then reaches shell state at priority 3, and the Claude Code TUI isn't a "running command" as far as shell integration is concerned — so the row renders **Idle** mid-turn.

The per-session sub-line directly beneath it reads "Running…" at that same moment, from the same store, because `sessionDetailText` already consults it. So the data is there and correct; it's the workspace-level chain that doesn't reach it.

Priority 1.5 closes it: below the tool label, which says more ("Reading file…"), and above everything after it, so the agent's own claim outranks the decaying inferences wmux makes about it.

**Why this commit is ~280 lines and not 3.** The status *text* chain and the status *colour* chain are written out twice in `WorkspaceRow.tsx`, and the single-session case sits in both — which is why this shows up as a colour mismatch about as often as a wrong word. I pulled both into `src/renderer/components/Sidebar/workspace-status.ts` and fed them from one shared input object, with the invariant written down. Side by side in one file, a priority added to one is visibly missing from the other. It's also the part worth testing and it couldn't be tested inside the component; 15 tests now pin it, including that the two chains agree.

Pure extraction otherwise — no behaviour change beyond priority 1.5. `WorkspaceRow.tsx` gets ~110 lines shorter as a result.

If you'd rather keep the chains inline, the behavioural part is a few lines and I can resubmit it that way.

## 2. Three edge cases in the hook bridge

**The per-tool `PostToolUse` hooks don't reach agent state.** `handleHookEvent` gates on `params.event`, and the `PostToolUse` entries `claude-context.ts` installs invoke the hook helper by bare tool name, which sends `{ tool }` with no `event` field. Those get dropped before reaching the bridge, so `case 'PostToolUse'` in `agent-hook-bridge.ts` isn't currently reached.

Since 0.48.0 that no longer costs the run itself — matcher-less `PreToolUse` declares it at the start of every tool, which is the important half. It still costs the other end: `PostToolUse` is the only report that arrives between approving a permission prompt and the turn closing, and it carries the freshness stamp that keeps a long turn inside the trust window. Without it, a turn spent in tracked tools ages out of `working` on the clock alone.

I handled it in the main process rather than by teaching the helper to send `event`, since hooks are written into `settings.json` once at install — the bare-tool form is already out in every existing config, and will keep arriving from helper vintages this build doesn't control. Open to the other direction if you'd prefer the helper be the source of truth.

**The 60s idle nudge reads as "Needs you".** `Notification` covers two unrelated things: a permission/question prompt, and the ~60s "Claude is waiting for your input" nudge armed when a turn *ends*. Parking the pane for both means a pane nobody has touched starts asking for attention a minute after it goes quiet. I hit it that way: a session was `/clear`'d and left alone, and 60s after its `Stop` the sidebar read **Needs you** with an empty conversation and nothing to answer.

Told apart by **state, not by message text**. Text-sniffing would break the day Claude Code rewords a prompt, and it would fail in the dangerous direction. A prompt can only occur inside a live turn, and `UserPromptSubmit`/`PreToolUse` both declare depth 1 before Claude Code can ask anything — so depth 0 means the turn is over.

Gated twice against reading too much into a default: on having seen an opening hook fire at all (a pre-0.48.0 `settings.json` sits at depth 0 through a whole turn), and on the pane actually having a record (a main process restarted mid-turn has none). In either of those cases it parks the pane, exactly as today.

That gating is also what keeps `noteHumanInput`'s bargain sound. It clears a block when the human types, and leans on "if the keystroke didn't satisfy the agent, the 60-second nudge puts the pane back to asking" — which works as long as the nudge doesn't also assert a block on a pane that has finished. The nudge still re-asserts inside a live turn, which is the case self-heal needs, and no longer outside one.

**Escape can leave a run asserted.** `noteHumanInput` treats a bare Escape as answering, so it clears `blocked` — but an interrupt mid-run leaves `working` set for a turn the user cancelled. Claude Code fires no hook on interrupt: `updatedAt` is byte-identical before and after, so there's no report to act on rather than a stale one. The pane keeps claiming `working` until the next prompt or the 15-minute trust window.

I matched on the single ESC byte exactly, rather than reusing `isAnsweringInput`'s "contains a bare ESC": Alt+key also arrives as ESC followed by a character, and reading that as an interrupt would retract a run nobody stopped. It only ever *clears*, and only for a pane that has declared something — so a plain shell is untouched, and a wrong guess under-reports for exactly one event.

## 3. Guard against a trailing `SubagentStop`

`SubagentStop` maps to an unconditional `runDepth: 1`, which is right when Claude Code fires the parent's `Stop` only after every subagent has stopped.

Traced at the hook helper on two independent panes here, the wire order came out the other way round: `Stop` first, `SubagentStop` **1.8–2.2s later**. In that order the trailing report re-asserts the run `Stop` had just ended, and the pane stays on **Running** — with the agent idle at its prompt — until the 15-minute trust window expires.

I'd expect this ordering to vary by Claude Code version and by how the subagent exits, so I wrote it as a **guard on top of the existing mapping rather than a replacement for it**: when `Stop` hasn't fired, behaviour is exactly as before. A subagent finishing may *sustain* a turn; it may never *start* one, and the current depth is the fact that separates the two.

I looked at whether the wall-clock gate in `agent-state.ts` could cover it, and I don't think it should. That gate drops reports *older* than what's been accepted, which is the right defence against two hook processes racing to the pipe — and this isn't a race, the trailing report is genuinely newer.

This also covers a second symptom that follows from the re-assertion: with depth back at 1, the 60s idle `Notification` no longer qualified as a nudge under (2), so a pane with nothing to answer went on to claim **Needs you** a minute later.

## 4. Keep a live question while background work runs

This is the one that finally explained my original report to me. An agent that starts a background shell and then asks the user a question shows **Running…** with the question still on screen, and nothing corrects it until the turn ends.

`PreToolUse`, `PostToolUse` and `SubagentStop` each report `awaitingHuman: false` alongside their run depth. For `PostToolUse` the reasoning in the comment is sound as written — a pending permission dialog would have stopped the tool from running at all — and it holds for an agent doing one thing at a time. It comes apart for the two things agents do most: a background shell and a parallel subagent both inherit the pane's `WMUX_SURFACE_ID`, so their tool lifecycle reports to the surface that's asking, and each event clears the block on the way past. Nothing re-asserts it, because the 60s idle `Notification` is only armed after a turn *ends*, and the turn is still open.

Those three events now report `runDepth` only. `runDepth` and `awaitingHuman` are independent facts, and the tool lifecycle speaks to the first. A block is ended by `UserPromptSubmit`, `Stop`, a keystroke in the pane (`noteHumanInput`), or an explicit report — the things that know a human dealt with it.

`resolveState` already ranks `blocked` above `runDepth > 0`, so a pane reads "needs you" **while** its background work keeps running, which is the state my report was describing.

**One trade-off I want to put in front of you rather than bury:** a prompt answered through `wmux answer-agent` (a relay, not a local keystroke) keeps saying "needs you" until the turn ends, rather than until the next tool. That's `answerAgent`'s rule 3 working as designed — answering never clears `blocked` — and `answeredAt` is what the UI shows in the meantime. I think it errs on the right side, but it *is* a change on the relay path, and it's your call. Happy to drop this commit and ship the other three if you'd rather keep the current behaviour there.

---

## Verification

- `npx vitest run` → **915 passed, 78 files** (881 in 77 on `master`). 34 tests added.
- `npm run build:main` clean.
- `npm run lint` → 17 errors / 15 warnings, **identical to unmodified `master`**. All pre-existing — I left them alone rather than mix a lint sweep into this.
- Rebased onto `master` at 0.50.0 (`8122b69`), no conflicts.

Live end-to-end, on a build of this branch: started two background shells, then asked a question. Polled `wmux agent-state` 300×: 72 samples `working`, then 228 samples `blocked "Claude needs your permission"`, with `runDepth` staying `1` throughout the blocked stretch, returning to `working` only after I answered. Sidebar read **Needs you** the whole time. That's (4).

(3) was traced at the hook helper across two panes; (2)'s Escape case by comparing `updatedAt` byte-for-byte before and after an interrupt.

## Not in this PR

- **No version bump.** `package.json` and `package-lock.json` are untouched.
- **No renderer test harness.** The 15 status tests cover the extracted chains as pure functions; the component wiring isn't tested, because vitest runs `environment: 'node'` with no jsdom here and adding one isn't a bug fix. Glad to do it as its own PR if you want that pinned down.
- **A notification-ring issue** I found while testing this — the attention ring not clearing when you type into the pane it's pointing at. Unrelated to agent state and shares no files, so it's a separate PR rather than a passenger here.

Thanks for 0.48.0 and for the detailed write-up on the issue — that's what let me go looking in the right places. Happy to rework any of this to fit how you'd rather it be shaped.

### DIFF
--- a/src/main/agent-hook-bridge.ts
+++ b/src/main/agent-hook-bridge.ts
@@ -195,9 +195,8 @@ export function hookToAgentReport(
     case 'PostToolUse':
       return { awaitingHuman: false, runDepth: 1 };
 
-    // One parallel subagent finished — and the parent turn is, by construction,
-    // still running: it has a result to read, and Claude Code fires the parent's
-    // `Stop` after every subagent has stopped.
+    // One parallel subagent finished. It may SUSTAIN a turn; it may never START
+    // one.
     //
     // This used to decrement the refcount, which was wrong in the way that
     // produced the loudest half of issue #151. Subagent hooks inherit the
@@ -207,7 +206,37 @@ export function hookToAgentReport(
     // with the parent and two siblings still working. A refcount only survives
     // contact with reality when every increment is paired, and these are not
     // paired; the parent's `Stop` is the pairing, and it is already decisive.
+    //
+    // Fixing that by asserting an unconditional depth of 1 rested on a premise
+    // that measurement does not support: that Claude Code fires the parent's
+    // `Stop` only after every subagent has stopped. It does not. Traced at the
+    // hook helper on two independent panes, the wire order is `Stop` first and
+    // `SubagentStop` 1.8-2.2s LATER — so the unconditional form resurrected the
+    // very run `Stop` had just ended, and the pane sat on "Running" until the
+    // 15-minute trust window with the agent idle at its prompt.
+    //
+    // The wall-clock gate in agent-state.ts cannot catch it, and that is the
+    // point worth remembering: it drops reports that are OLDER than what has
+    // been accepted, which is the right defence against two hook processes
+    // racing to the pipe. This is not a race. The trailing report is genuinely
+    // newer; it is the EVENT ORDER that is counter-intuitive, and no amount of
+    // timestamp discipline fixes a report that is late by design.
+    //
+    // Reading the current depth is what separates the two situations, using the
+    // one fact that distinguishes them: the parent's `Stop` has already zeroed
+    // it. So a subagent finishing inside a live turn holds that turn open, and
+    // a subagent finishing after the turn ended says nothing at all. Declining
+    // to report also leaves the freshness stamp alone, so the 60s idle
+    // `Notification` that follows still sees depth 0 and is correctly read as a
+    // nudge rather than a prompt — the resurrection defeated that gate too, and
+    // the pane went on to claim "Needs you" with nothing to answer.
+    //
+    // `known` is checked for the same reason `Notification` checks it: a pane
+    // wmux holds no record for has a depth of 0 by DEFAULT, not by observation.
+    // Nothing is lost by staying quiet there — a turn wmux never saw begin is
+    // not one a subagent's completion should invent.
     case 'SubagentStop':
+      if (!ctx.known || ctx.runDepth === 0) return null;
       return { awaitingHuman: false, runDepth: 1 };
 
     // The turn is over: nothing can still be running and nothing can still be
--- a/src/main/agent-hook-bridge.ts
+++ b/src/main/agent-hook-bridge.ts
@@ -21,7 +21,7 @@
  */
 
 import { SurfaceId } from '../shared/types';
-import { reportAgent, releaseAgent, ReportAgentParams } from './agent-state';
+import { getAgentState, reportAgent, releaseAgent, ReportAgentParams } from './agent-state';
 
 /** The hook events wmux registers. */
 export type ClaudeHookEvent =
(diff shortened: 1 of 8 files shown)


amirlehmam:
Merged, all four commits, in **1.0.0**.

This is the best PR this project has had. Not because the diff is large — because every claim in it is one you measured, and two of the four commits turn on a fact I had asserted in a comment and never checked.

## On each commit

**1 — the single-session path.** You are right that the ~280 lines are justified, and the reason is stronger than "it's testable". The two chains had *already* diverged on exactly the line you were fixing: `workingSessions` was consulted only under `sessionCount >= 2` in both, so one working pane rendered the word "Idle" in the `--done` colour. Two implementations of one precedence order is how that happens, and it will happen again. I would not have taken "extract it and add a test" on its own; I take it because the file now makes the next divergence visible on one screen. Keep the extraction.

**2 — the bare-tool `PostToolUse` form.** This one is mine and it is embarrassing: I wrote `case 'PostToolUse'` in a switch that `handleHookEvent` could never reach it through, and the 0.48.0 write-up describes its behaviour as if it were live. You resolved it in the main process rather than in the helper, and that is the right call for the reason you gave — the bare-tool form is already out in every existing `settings.json` and will keep arriving from helper vintages this build does not control. A fix that requires the other side to be current is not a fix.

**The `Notification` discriminator is the part I would have got wrong.** Using `runDepth === 0` rather than the message text is correct, and the two gates are what make it safe rather than clever: `turnStartTracked` because a pre-0.48.0 config sits at depth 0 through a whole turn, and `known` because a depth of 0 by *default* is not a depth of 0 by *observation*. Both fail toward parking the pane. That is the direction that costs a false "Needs you" instead of swallowing a real permission prompt, and getting that asymmetry right is the whole job here.

Making `turnStartTracked` global rather than per-surface deserves its comment — I read it as a bug for about ten seconds before the comment explained that the question is "does settings.json carry the 0.48.0 hook set", which is one file for every pane.

**3 — the trailing `SubagentStop`.** I asserted in the 0.48.0 comment that Claude Code fires the parent's `Stop` only after every subagent has stopped. You measured the opposite, on two independent panes, with a 1.8–2.2 s gap. I had no evidence for that claim and it should not have been written as one.

Shipping it as a guard on top of the existing mapping rather than a replacement is the right shape given the ordering may vary by version and by how the subagent exits. And your note that the wall-clock gate cannot cover it is the part worth keeping in mind: that gate drops reports *older* than what was accepted, which is the right defence against two hook processes racing to the pipe — this is not a race, the trailing report is genuinely newer, and no amount of timestamp discipline fixes a report that is late by design.

**4 — keeping a live question while background work runs. Taken, not dropped.**

You flagged the `answer-agent` trade-off and offered to drop this one. I want it, and I think your instinct about which side to err on is right.

The premise I wrote — "a pending permission dialog would have stopped the tool from running at all" — holds for an agent doing one thing at a time and is false for the two things agents actually do. Background shells and parallel subagents inherit the pane's `WMUX_SURFACE_ID`, so their tool lifecycle reports to the surface that is asking, and each event retracted the flag on the way past with nothing to re-assert it. "Blocked AND busy" is a legitimate state and `resolveState` already renders it correctly; the events were lying about the first half.

On the relay path specifically: a block that lingers until the turn ends is visible and self-correcting, and `answeredAt` says "sent — waiting for the agent" in the meantime. A block that vanishes while the agent is still stuck is the failure this whole subsystem exists to prevent. That is rule 3 working, not a regression against it.

## What I did not change

Nothing. Merged as submitted. `npx vitest run` is at 1004 passing on master with the rest of 1.0.0 on top, and lint is *below* the 32-problem baseline you measured, because deduplication elsewhere in the release removed some of it.

## One thing your PR made me find

Not yours, but it came out of reading the 60-second-nudge reasoning: `stripWmuxBlock` and the block splice existed in two copies, and when I added a third agent (#165) the new copy silently dropped a `trimEnd()` the others depended on — appending one blank line per launch to a file the agent loads every session. Same failure mode as your commit 1, in a different file. They share one implementation now.

Thank you for running this as your daily driver for two weeks and writing down what you saw. The edge cases you listed as still-hard — nested subagents, a turn ending while a subagent runs, a session answered from another pane — are exactly the ones I would most like eyes on next, and you now have more evidence about them than I do.

PR #161, opened 2026-08-13
fix(security): use safer defaults for agent integrations

Marv-In-Make:
## Summary

- pin `chrome-devtools-mcp` to `1.7.0` instead of persisting the mutable `@latest` npm tag
- keep Claude's normal permission prompts for orchestrated workers by default
- retain `--dangerously-skip-permissions` behind the explicit `WMUX_ORCHESTRATOR_SKIP_PERMISSIONS=1` opt-in
- disclose all eight Claude hook families and the OpenCode plugin path in the integration consent dialog
- document the opt-in and add regression coverage for these defaults

## Why

The custom MCP command is stored in the user's global Claude settings and may run long after the wmux release that created it. Using `npx -y ...@latest` means future package contents can be downloaded and executed without a corresponding wmux code change.

The orchestrator also passed `--dangerously-skip-permissions` to every Claude worker. That removes the user confirmation boundary for shell commands and file access, including when an agent processes an untrusted repository or prompt.

This change preserves both capabilities while making the safer behavior the default.

## User impact

- Existing browser integration entries managed by wmux migrate from `@latest` to the pinned package version when integration settings are reconciled.
- Orchestrated Claude workers may ask for tool approval. Users who accept the risk for a trusted task can restore the former behavior by setting `WMUX_ORCHESTRATOR_SKIP_PERMISSIONS=1` before starting wmux.
- The first-run consent dialog now accurately lists every hook family and plugin file wmux installs.

## Validation

- `npm run typecheck` — passed
- `npm test -- --run tests/unit/agent-integration.test.ts` — 23/23 passed
- `node --check resources/wmux-orchestrator/scripts/launch-agent.js` — passed
- full `npm test -- --reporter=dot` — 880 passed, 5 failed

The five full-suite failures are all in `tests/unit/orchestration-status-vocab.test.ts` and reproduce the existing Bash path failure on this Windows environment (`bash: : No such file or directory`). They are unrelated to the files changed here.

## Out of scope

CDP authentication, dynamic CDP port propagation, updater signing enforcement and release-workflow hardening should remain separate changes because they require broader architectural or release-policy decisions.

### DIFF
--- a/resources/wmux-orchestrator/README.md
+++ b/resources/wmux-orchestrator/README.md
@@ -55,6 +55,21 @@ Run it:
 /wmux:orchestrate "Refactor the auth system to use JWT tokens"
 ```
 
+Claude workers keep their normal tool permission prompts by default. To bypass
+them for a trusted task, opt in explicitly before starting wmux:
+
+```powershell
+[Environment]::SetEnvironmentVariable('WMUX_ORCHESTRATOR_SKIP_PERMISSIONS', '1', 'User')
+```
+
+This makes orchestrated Claude workers start with
+`--dangerously-skip-permissions`. Remove the variable to restore the safe
+default:
+
+```powershell
+[Environment]::SetEnvironmentVariable('WMUX_ORCHESTRATOR_SKIP_PERMISSIONS', $null, 'User')
+```
+
 What happens next:
 
 1. The orchestrator analyzes your codebase and builds a wave plan
--- a/resources/wmux-orchestrator/scripts/launch-agent.js
+++ b/resources/wmux-orchestrator/scripts/launch-agent.js
@@ -10,36 +10,46 @@
 const { execFileSync } = require('child_process');
 const fs = require('fs');
 
-const promptFile = process.argv[2];
-if (!promptFile) {
-  console.error('Usage: node launch-agent.js <prompt-file>');
-  process.exit(1);
-}
-
-if (!fs.existsSync(promptFile)) {
-  console.error(`Prompt file not found: ${promptFile}`);
-  process.exit(1);
-}
-
-const prompt = fs.readFileSync(promptFile, 'utf8');
-
-const agentCmd = (process.env.WMUX_AGENT_CMD || 'claude').toLowerCase();
-
-try {
-  if (agentCmd === 'opencode') {
-    // opencode run streams formatted progress; the user can watch.
-    // '--' stops flag parsing from consuming the prompt.
-    execFileSync('opencode', ['run', '--', prompt], { stdio: 'inherit' });
-  } else {
-    // --dangerously-skip-permissions: auto-approve all tools (interactive mode)
-    // '--' stops Commander.js variadic flags from consuming the prompt
-    // NOTE: do NOT use --bare — it skips keychain/OAuth and causes "Not logged in"
-    execFileSync('claude', [
-      '--dangerously-skip-permissions',
-      '--',
-      prompt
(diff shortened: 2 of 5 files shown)


amirlehmam:
Merged in **1.0.0**, with one change on top that I want to put in front of you rather than bury in a commit.

All three changes are right, and 1.0.0 is the correct release to make safe the default in.

## What I took as-is

**Pinning `chrome-devtools-mcp`.** The argument is the one that matters and it is not hypothetical: the command is persisted in the user's *global* Claude settings and executes on every session that loads them, long after the wmux release that wrote it is gone. `npx -y ...@latest` means the contents of that package can change under a user who never updated anything. Pinning is the only version of this that is defensible.

**The permission default.** This is the one I went back and forth on, because a wave of six workers that each stop on their first `Bash` call is a real cost to the feature. I landed where you did: an orchestrated worker analyses a codebase it was pointed at, and repository content is model input, so "the user asked for this task" is not the same as "the user vouched for everything the agent will read". The capability survives behind the env var, which is what makes it a default rather than a removal.

**The consent disclosure.** It had silently drifted from four hooks to eight since 0.48.0, which makes the dialog *wrong* rather than merely incomplete — it is the one place a user is asked to agree to something. Good catch, and moving it to an exported constant so a test can assert every hook family is named is the right way to stop it drifting again.

## What I changed

`ensureChromeDevtoolsConfig` now asks *"did wmux write this entry"* rather than *"is this entry what wmux wants"*.

The PR's version rewrites whenever `JSON.stringify(existing) !== JSON.stringify(desired)`. That does migrate every existing `@latest` install, which is the goal — but it also means any entry the user retuned by hand gets clobbered **on every launch**, silently, forever. And it puts the write path in direct contradiction with `removeChromeDevtoolsConfig` a few lines below, which already documents the opposite care:

> Only an entry that points at wmux's own CDP proxy port is removed — a user who has since pointed `chrome-devtools` somewhere of their own keeps it.

Uninstall respecting the user's edit while install stomps it is the exact complaint in #132. `isWmuxAuthoredMcpEntry` matches wmux's signature instead — npx, `chrome-devtools-mcp` at any pin, aimed at wmux's own proxy port — so the `@latest` → pinned migration still reaches every existing install, and a different port, a global install, or someone else's server is left alone. It also survives the next release that moves the pin, which the equality check would not have.

Two smaller ones in the same area:

- **`launch-agent.js` resolves the agent to an absolute path** instead of letting the OS resolve a bare name against whatever `PATH` the pane inherited, where a writable directory earlier in the order shadows it. That matters most in this exact script, whose job is to hand a coding agent a prompt and let it edit the repo. It also turns a bare `ENOENT` — which reads as "the agent crashed" — into a message naming what is missing.
- **The worker prints one line** saying why it is about to ask for permission and how to opt out. A silent change of this default reads as a bug: a wave that used to run unattended starts stopping, with nothing on screen connecting that to a decision anyone made. It goes to the worker's own pane, which is where the prompts appear.

## On your "out of scope" list

Agreed on all four, and thank you for naming them rather than half-doing one. CDP authentication in particular is a real gap that deserves its own issue and its own argument — the proxy currently leans on a loopback bind plus the host/origin guards, and "is that enough" is not a question to answer inside a PR about npm pins.

## Note on your test run

The five `orchestration-status-vocab.test.ts` failures were environmental as you suspected — bash resolution on Windows. Full suite is green here at 1004 passing.

Thanks for this. Security PRs that preserve the capability behind an explicit opt-in, instead of removing it and calling it hardening, are rarer than they should be.

PR #163, opened 2026-08-14
fix(terminal): route OSC 8 hyperlinks through wmux

### LINKED ISSUE #162: bug: OSC 8 terminal hyperlinks do not open
## Summary

OSC 8 hyperlinks rendered in wmux terminals show xterm's confirmation dialog, but accepting it does not open the URL. Plain-text URLs in the same terminal open correctly.

This affects hyperlinks emitted by terminal applications such as coding-agent TUIs, including when those applications run inside a terminal multiplexer.

## Reproduction

In a wmux terminal, run the following in PowerShell:

```powershell
$e = [char]27
Write-Host "PLAIN: https://example.com"
Write-Host "OSC8: $e]8;;https://example.com$e\CLICK-ME$e]8;;$e\"
```

1. Click the plain-text URL: it opens in the wmux browser.
2. Click `CLICK-ME`: xterm displays its external-link warning.
3. Accept the warning: nothing opens.

The same failure occurs for OSC 8 links emitted by agent TUIs. It is independent of SSH, terminal mouse capture, and the application producing the link.

## Expected behavior

OSC 8 and plain-text HTTP(S) links should use the same wmux link-opening behavior:

- normal click opens in the wmux browser
- Ctrl/Meta-click opens externally

## Root cause

`WebLinksAddon` is given a custom callback for detected plain-text URLs, but the xterm `Terminal` is not given a `linkHandler` option.

OSC 8 links are handled by xterm's OSC link provider rather than `WebLinksAddon`. Without `Terminal.options.linkHandler`, they fall back to xterm's default confirmation + `window.open()` path. In Electron that call does not perform wmux's browser routing, so navigation is dropped.

## Proposed fix

Provide `Terminal.options.linkHandler.activate` and route it through the same HTTP(S)-validated `openInWmuxBrowser` callback used by `WebLinksAddon`.

## Verification

Tested locally on Windows in both development and packaged production builds:

- plain-text URL opens in wmux
- OSC 8 URL opens in wmux
- Ctrl-click opens externally
- OSC 8 links emitted inside herdr-hosted Codex/Claude sessions open correctly
- terminal mouse interaction remains functional in the packaged build

Unit tests, the full Vitest suite, TypeScript checks, and renderer/main production builds pass.

LevSky22:
## Summary

- route xterm OSC 8 hyperlinks through wmux's browser-opening path
- share the same activation callback with `WebLinksAddon` so detected and explicit terminal links behave consistently
- accept only HTTP(S) targets and preserve Ctrl/Meta-click external opening
- add focused unit coverage for routing, modifiers, and unsafe targets

Fixes #162.

## Root cause

Plain-text URLs are detected by `WebLinksAddon`, which already receives a wmux-specific callback. OSC 8 hyperlinks take xterm's separate OSC link-provider path and read `Terminal.options.linkHandler` instead.

Because wmux did not configure `linkHandler`, OSC 8 activation fell back to xterm's confirmation and `window.open()` behavior. That fallback does not route through wmux's Electron/browser handling, so accepting the dialog did not navigate anywhere.

## User impact

Links emitted by coding agents and other terminal applications now behave like ordinary detected URLs:

- normal click opens in the wmux browser
- Ctrl/Meta-click opens externally
- unsupported or malformed schemes are ignored

## Validation

- `npm test` — 884 passed, 5 skipped
- `npm run typecheck`
- `npm run build:main`
- `npm run build:renderer`
- `npm exec -- electron-builder --dir`
- packaged Windows manual test:
  - plain-text and OSC 8 PowerShell links open
  - agent-emitted OSC 8 links open from Codex/Claude running inside herdr
  - herdr terminal mouse interaction remains functional

`npm run lint` still reports the repository's existing baseline findings (17 errors and 15 warnings); the new helper passes focused linting and the modified lines introduce no new lint finding.

### DIFF
--- a/src/renderer/hooks/useTerminal.ts
+++ b/src/renderer/hooks/useTerminal.ts
@@ -12,7 +12,7 @@ import { useT } from '../i18n';
 import { collectActiveTerminalSurfaceIds } from '../store/split-utils';
 import { SplitNode, ThemeConfig } from '../../shared/types';
 import { UserColorScheme } from '../store/settings-slice';
-import { openInWmuxBrowser } from '../utils/open-in-browser';
+import { activateTerminalLink, terminalLinkHandler } from '../utils/terminal-links';
 import { attachVisibleRenderer, RendererHandle } from '../utils/terminal-renderer';
 import { trimTrailingWhitespace } from '../utils/copy-text';
 import { handleShiftEnter, isShiftEnter } from './terminal-keys';
@@ -392,6 +392,7 @@ export function useTerminal({ surfaceId, shell, cwd, visible = true, focused = t
       // enabled, or toggling it would require recreating every terminal.
       allowTransparency: true,
       allowProposedApi: true,
+      linkHandler: terminalLinkHandler,
       scrollback: prefs.scrollbackLines || 10000,
     });
 
@@ -408,10 +409,7 @@ export function useTerminal({ surfaceId, shell, cwd, visible = true, focused = t
 
     // Create and load addons
     const fitAddon = new FitAddon();
-    const webLinksAddon = new WebLinksAddon((event, uri) => {
-      const forceExternal = !!(event as MouseEvent)?.ctrlKey || !!(event as MouseEvent)?.metaKey;
-      openInWmuxBrowser(uri, { forceExternal });
-    });
+    const webLinksAddon = new WebLinksAddon(activateTerminalLink);
     const searchAddon = new SearchAddon();
     const unicode11Addon = new Unicode11Addon();
     const imageAddon = new ImageAddon();
--- a/src/renderer/utils/terminal-links.ts
+++ b/src/renderer/utils/terminal-links.ts
@@ -1,0 +1,23 @@
+import type { ILinkHandler } from '@xterm/xterm';
+import { openInWmuxBrowser } from './open-in-browser';
+
+/** Route terminal links through wmux instead of xterm's window.open fallback. */
+export function activateTerminalLink(event: MouseEvent, uri: string): void {
+  let protocol: string;
+  try {
+    protocol = new URL(uri).protocol;
+  } catch {
+    return;
+  }
+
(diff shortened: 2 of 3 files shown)


amirlehmam:
Merged in **1.0.0**, closes #162.

The diagnosis is exactly right and it is the kind that is hard to reach from the outside: xterm has *two* link paths, and they are configured in two different places. `WebLinksAddon` takes a callback in its constructor and handles detected plain-text URLs; OSC 8 hyperlinks go through the OSC link provider, which reads `Terminal.options.linkHandler` and nothing else. wmux configured the first and not the second, so explicit hyperlinks fell back to xterm's confirm-then-`window.open()` — and under Electron that call does not route through wmux's browser handling, so accepting the dialog navigated nowhere.

What makes it a nasty report to receive is that the dialog *appears*. Everything looks wired up right up to the last step, which is why "OSC 8 links do nothing" sat unnoticed while plain URLs worked fine in the same terminal.

Two things I'd have been sloppier about:

**Sharing one activation function rather than two callbacks that agree.** `activateTerminalLink` is used by both paths, so detected and explicit links cannot drift apart in modifier handling or scheme policy. Your test asserting `terminalLinkHandler.activate === activateTerminalLink` pins that directly, which is a nice touch — it fails if someone later "helpfully" wraps one of them.

**Validating the scheme.** Providing a `linkHandler` replaces xterm's default confirmation, so wmux now owns that decision. `javascript:`, `data:` and `file:` targets being dropped matters more once the guard rail is gone than it did before, and the table-driven test covers the cases that would actually be attempted.

Verification was thorough — packaged build, not just dev, and you checked that terminal mouse interaction still worked, which is the thing most likely to break as collateral when touching terminal options. That turned out to be worth checking for an unrelated reason: #164 landed in this same release and touches mouse state on the remount path.

Thanks — clean, narrow, tested, and it fixed the issue you filed rather than the one that was easy to fix.

Issue #134, opened 2026-07-30
another issue with save session?


arnonuem:
It seems that the absolute paths of the open terminal windows are not stored or restored.
I manually saved a session which had two terminals open with a path like D:\... but after restore it was C:\....

Since i use git work trees which sit on different disks and which are tied to wmux sessions that leaves a chaos behind because i need to find out what the original path was when i need to restore a session (because of an update maybe).

Can you please investigate this?

amirlehmam:
Investigated, reproduced in the code, fixed in **v0.40.0**. Your instinct was right — the paths genuinely were not being stored.

## What was happening

A terminal in wmux carries two directories, and only one of them was ever persisted:

- `cwd` — where the terminal was **told to start**. A spawn argument.
- `currentCwd` — where shell integration **last saw it**, updated on every prompt.

Saved sessions serialised `cwd`. For a tab you open the ordinary way, `cwd` is `undefined` — nothing told it where to start. So on restore every terminal fell through to a single, workspace-level `cwd`... which is itself written by whichever pane reported a prompt most recently. Two terminals, one workspace, one value: last writer wins.

That is why you saw `D:\` become `C:\`. The workspace value was either empty or belonged to the other pane, and with nothing usable the shell fell back to its own default — your home directory, on C:.

Worth being clear that this was not a rounding error. Every terminal in a workspace came back in the same place, and for your setup — worktrees on separate drives, one session per worktree — the directory *is* the identity of the session. Losing it does not degrade the restore, it makes the restore pointless, which is exactly how you described it.

## The fix

A new `freezeSurfaceCwds()` promotes each surface's live directory into its persisted one, per tab, for the copy of the tree that goes into the saved session. Both paths use it — the named session you save by hand, and the 30-second auto-save that survives an update.

Two design points, since they are the reason this is safe:

**It happens at save time, not on every prompt.** `cwd` is a spawn argument — a fact about how the terminal was born — and rewriting it live would be editing a running terminal's history. It is also passed as a prop into the hook that creates PTYs, so a value that changes every time you `cd` is asking for trouble.

**It freezes into a copy.** The live store keeps its own `cwd` untouched; only the snapshot collapses the two facts into one, because "where should this reopen?" is the only question a snapshot needs to answer. There is a test pinning exactly that invariant, alongside ones for two panes on different drives staying on different drives, background tabs being frozen too (not just the visible one), and a terminal that has `cd`-ed away from its start directory persisting where it *is* rather than where it began.

[...]

Issue #158, opened 2026-08-13
`wmux ping` cannot distinguish "not running" from "not on this session's PATH", and the agent block tells the reader to assume the first


capad-xyz:
Follow-on from #152, and I think it undercuts that fix rather than contradicting it. The block is right to make the reader check instead of assume. The problem is what the check can actually tell them.

## The conflation

`resources/claude-instructions.md:18`:

> **If it does not answer** — command not found, no reply, or an error — then wmux
> is not running for this session. Ignore the whole section and use your normal
> tools.

Three symptoms, one conclusion. But they are not the same fact:

- **no reply on the pipe**: wmux is not running. Ignoring the section is correct.
- **command not found**: says nothing about whether wmux is running. It says this session cannot reach the CLI.

The block instructs the reader to treat the second as the first.

## Why the second case is the common one, not the exotic one

Verified from source: `src/main/pty-manager.ts:323-341` prepends the shim dirs to the environment of PTYs **wmux spawns**, with the comment "Make bare `wmux` resolvable in every spawned shell AND all its children". That is the only place PATH is touched. There is no `setx`, no registry write, nothing that persists PATH anywhere in `src/`.

So `wmux` resolves inside a pane and its children, and nowhere else unless the user put it on PATH by hand. Every session wmux did not spawn - the Claude Code desktop app, a plain terminal, an SSH session, a scheduled run, a session that was already open before wmux started - has no `wmux` on PATH by default.

That is precisely the population #152 rewrote the block for. The block loads globally so it can reach sessions that are not inside wmux, and then hands those sessions a probe that, for most of them, fails for a reason the block does not account for.

## Live, on this machine

Verified: wmux 0.50.0 running (20 processes), `pid 28680` listening on `127.0.0.1:9222`, named pipe answering. And:

```
$ wmux ping
bash: wmux: command not found

PS> wmux ping
wmux : The term 'wmux' is not recognized as a name of a cmdlet, function, script file, or executable program.
```

Following the block, every agent on this machine concludes wmux is absent and stops using the browser panel. wmux is fine. Only the entry point is missing.

[...]

Issue #145, opened 2026-08-05
Autobackup does its own thing


arnonuem:
Why does autobackup and autorestore behave differently from manual save and load?

When i save my session (before an update of wmux - and thanks god i did) it correctly saves the workdir paths of the sessions and the browser states per session. When i load this session after update all is now correctly restored.

But before update there is an autobackup now and after update an autorestore?

Either the autorestore is not triggered automatically or the autorestore does not restore the same things like manual restore.

At least browser-states are back to default (wmux github page).

I was on version 0.41.x - might be fixed now.

amirlehmam:
You diagnosed this correctly — "the autorestore does not restore the same things like manual restore" is exactly what was happening, and it wasn't fixed in 0.41.x. Thanks for saving your session before updating, and for reporting it.

**Root cause.** On the first launch of a new version, wmux archives the volatile auto-session (`session.json`, whose PTYs died with the old process) as a *named* session — "Auto-backup vX.Y.Z" — and then clears it. That backup is what the post-update startup restores, since there's no auto-session left to come back to. It's the safety net added in #113.

The conversion was lossy. `backupAutoSession()` copied 5 of the 8 fields the auto-save actually writes:

```js
workspaces: win.workspaces.map(w => ({
  title, customColor, shell, cwd, splitTree,   // ← that's all it kept
})),
```

So `browserUrl` was dropped on the way in, and every workspace's browser came back on the default page. `browserWidth` and `pinned` went the same way, and it only ever read `windows[0]` — with two windows open, the second window's workspaces were gone entirely.

Your manual save was unaffected because that path (`handleSaveSession`) always wrote the full set. Same data, two writers, one of them incomplete — hence "does its own thing".

The reason it survived this long: `SessionData` never declared `browserUrl`/`browserWidth` even though the renderer has been writing them since the auto-save existed, so the type was a *subset* of the real payload and dropping fields looked correct to the compiler.

**Fixed in [0.44.0](https://github.com/amirlehmam/wmux/releases/tag/v0.44.0):**

- the update backup carries `browserUrl`, `browserWidth` and `pinned`, and every window's workspaces rather than just the first (multi-window layouts come back flattened into one window — losing the split is a smaller surprise than losing the tabs);
- `SessionData` now declares those fields, so a future save path can't silently drop them again;
- a manual save keeps `pinned` too, which only the auto-save was doing;
[...]

PR #166, opened 2026-08-14
feat(devcontainer): drive wmux from a container over the TCP bridge

stlindenau:
## Summary

Seven commits on top of `master` (v0.50.0). The headline one lets the `wmux` CLI, the Claude Code hooks and the shell integration reach `\\.\pipe\wmux` from inside a Linux devcontainer; the rest are standalone fixes found along the way.

Each commit stands alone and explains its own reasoning in the message — reviewing commit by commit will be a lot easier than reading the combined diff. Happy to split this into separate PRs if you'd prefer; opened as a draft so you can say so before I do.

## What's in it

- **`feat(devcontainer)`** (`9fd08ed`) — `wmux bridge --wsl` runs the byte relay *inside* WSL2 and reaches the Windows pipe via [`npiperelay.exe`](https://github.com/albertony/npiperelay) over interop. `0.0.0.0` there is the WSL2 network namespace: reachable from containers through the Docker host gateway, not from the LAN, and no firewall rule needed. Builds on the remote transport from #78. Adds `wmux raw-v1` so the bash integration can send V1 lines through the CLI's transport instead of a Windows temp file it cannot write, plus `report_startup_command` so a restored pane re-enters its container instead of coming up as a bare WSL prompt. AF_VSOCK, a Windows-side listener and cross-boundary Unix sockets were all tried first and are documented as rejected in the code. Setup guide: `docs/DEVCONTAINER.md`.
- **`fix(cwd)`** (`f5f782c`) — a workspace tracks `posixCwd` separately from `cwd`, so a pwsh pane no longer rewrites the directory a neighbouring WSL pane falls back to. Extends what #134 did per-terminal to the workspace, and is additive: sessions saved before the field existed recover via an `isPosixPath(cwd)` seed.
- **`feat(config)`** (`0fa84ab`) — warn when `~/.wmux/config.toml` fails to read or parse. The whole file is discarded on failure, so one stray character presents as "my setting has no effect" with nothing pointing at the file. Deduped, because `loadUserConfig` runs on every WSL pane spawn.
- **`fix(cli)`** (`16b0d4c`) — `wmux config path` built the path from `os.homedir()` instead of asking the instance, so from WSL or a container it printed `/home/vscode\.wmux\config.toml` — not the file wmux reads, and not a well-formed path on either OS.
- **`fix(shell-integration)`** (`d1c1b58`) — `resources/shell-integration/wmux-bash-integration.sh` had not moved since v0.5.2 while the copy under `src/` had. Consumers that read it out of a repo checkout got the pre-bridge `_wmux_report` and an error on every prompt. Plain copy of the `src/` file, no hand-edits. The `.cmd` and `.ps1` copies have drifted the same way and are deliberately left for a separate change.
- **`fix(cli)`** (`c693c06`) — three places in the CLI that assumed one OS's path conventions. `findNpiperelay()` split `$PATH` on a hard-coded `:`, so on native Windows the whole search collapsed to one nonsense entry; it now splits on `path.delimiter`, drops empties, and reads `Path` as well as `PATH`. The `config path` and `locales path` fallbacks — the branch taken when no instance answers — chose their separator from the *host* rather than from the home directory actually in hand, producing `/home/vscode\.wmux\config.toml`; they now pick `path.posix.join` for a POSIX home and `path.join` otherwise.
- **`test(bash-path)`** (`64fb8d8`) — repairs the test fallout of the commit above. `c693c06` also made the `bashPathCandidates()` test helper platform-conditional (Git Bash's `/c` first on Windows, WSL's `/mnt/c` elsewhere) so the orchestrator suites can resolve a drive path under a Linux bash, but `tests/unit/bash-path.test.ts` still asserted the fixed Git-Bash-first order it was written against in #130 — green on Windows, four failures on Linux and macOS. The set of spellings offered is invariant and is now asserted order-independently; the order is not, and is asserted once against the same `process.platform` condition the helper uses. No behaviour change.

`resources/cli/wmux.js` and `resources/cli/wmux-hook.js` are the `tsc -p tsconfig.node.json` output of their sources — re-checked at this head, and byte-identical to a fresh `npm run build:main`, so there are no hand-edits in the artifacts.

## Testing

Measured on Linux (a devcontainer), which is not where most of this code lives, so the honest form is a comparison against `master` on the same machine. Baseline: `8122b69` checked out in a detached worktree, same `node_modules`, same commands.

|  | `master` `8122b69` | this branch `64fb8d8` |
|---|---|---|
| `npm run typecheck` | clean | clean |
| `npm run lint` | 32 problems (17 errors, 15 warnings) | identical — 32 (17 / 15) |
| `npx vitest run` | 6 failed, 865 passed, 2 skipped (873) | 6 failed, 909 passed, 6 skipped (921) |

The failing set is **identical to `master`, one for one** — same files, same test names:

```
tests/unit/orchestration-watcher.test.ts          (suite-level)
tests/unit/pty-ledger.test.ts        × 2          (live Win32_Process probe)
tests/unit/pty-manager.test.ts       × 1          (resolveSpawnCwd)
tests/unit/shell-context-menu.test.ts × 3         (directoryFromArgv)
```

All of them are Windows-path or live-Win32-probe cases that cannot pass off Windows; none of them are new. Nothing this branch adds breaks anything that worked before it.

The four new suites pass: `workspace-posix-cwd` 21/21, `wmux-hook-remote-transport` 5/5, `cli-config-path` 3/3, and `wmux-bridge-transport` 1/1 — including the case that relays TCP to a Unix-socket upstream, which is the transport shape the container path depends on. That suite's other four cases (npiperelay stdio, warm relay pool) are gated to WSL2 and skip here, which is where the +4 skipped comes from.

A run on Windows is still worth having before merge, and matters a little more now: `c693c06` and `64fb8d8` both branch on `process.platform`, so each has a side no Linux run exercises.

The devcontainer path itself is in daily use in the setup `docs/DEVCONTAINER.md` describes (Windows 11 + WSL2 + Docker devcontainer, Claude Code inside the container).

## Notes

- `wmux locales path` still cannot be resolved the way `wmux config path` now is — `config.get` carries no locales directory, so there is nothing to ask the instance for. `c693c06` only makes its local fallback produce a well-formed path for the OS it runs on; it can still be the wrong path when the CLI and wmux are on opposite sides of the WSL boundary.
- Deadlines gained a floor on the slow transports: the 5s default sat below the container→WSL2→pipe round-trip, so calls that had already succeeded were reported as timeouts.

---

🤖 This PR and the code in it were written with the help of [Claude Code](https://claude.com/claude-code); every commit but `c693c06`, where the trailer was dropped by accident, carries a `Co-Authored-By: Claude` trailer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -326,6 +326,12 @@ wmux --remote 127.0.0.1:9787 --token <TOKEN> list-workspaces
 wmux --remote 127.0.0.1:9787 --token <TOKEN> new-workspace --title "api"
 # Or set once: WMUX_REMOTE=127.0.0.1:9787 and WMUX_REMOTE_TOKEN=<TOKEN>
 
+# From a devcontainer — same transport, no SSH tunnel. Run the bridge in WSL2
+# (--wsl binds 0.0.0.0 there, reachable from the container, not from the LAN):
+wmux bridge --wsl                  # relays to \\.\pipe\wmux via npiperelay.exe
+# Then in the container: WMUX_REMOTE=host.docker.internal:9787 + WMUX_REMOTE_TOKEN
+# Full setup: docs/DEVCONTAINER.md
+
 # Agents
 wmux agent spawn --cmd "claude --resume abc" --label "Research"
 wmux agent spawn-batch --json '[{"cmd":"claude","label":"Agent 1"},{"cmd":"claude","label":"Agent 2"}]'
@@ -345,9 +351,14 @@ Connect to `\\.\pipe\wmux` for programmatic control. Two protocols supported:
 report_pwd <surface_id> <path>
 report_git_branch <surface_id> <branch> [dirty]
 report_shell_state <surface_id> idle|running|interrupted
+report_startup_command <surface_id> <command>   # how to restore this surface; must be cwd-independent
 notify <surface_id> <text>
 ping
 ```
+
+`wmux raw-v1 "<line>"` sends any of these through the CLI's transport, which is
+what lets a shell with no reachable pipe — inside a devcontainer — still report.
+See [docs/DEVCONTAINER.md](docs/DEVCONTAINER.md).
 
 **V2** (JSON-RPC, used by CLI and automation):
 ```json
--- a/docs/DEVCONTAINER.md
+++ b/docs/DEVCONTAINER.md
@@ -1,0 +1,256 @@
+# Driving wmux from a devcontainer
+
+wmux runs on Windows, but a Claude Code session often does not: it runs in a
+Linux devcontainer on top of WSL2 + Docker. Nothing in that container can open
+`\\.\pipe\wmux`, so the `wmux` CLI, the Claude Code hooks and the shell
+integration all fail silently — panes show no cwd or git branch, the sidebar
+never leaves "Running", and `wmux agent spawn` cannot reach the app at all.
+
+This guide sets up a path from the container to that pipe. It needs no Bosch
+feature, no vendor tooling and no code you have to write: one binary, one
+long-running command, and two environment variables.
+
+## How it fits together
+
+```
+  Devcontainer (Linux)                WSL2 distro                     Windows
+ ┌──────────────────────┐   TCP    ┌────────────────────┐  interop  ┌─────────┐
+ │ claude               │ ───────► │ wmux bridge --wsl  │ ────────► │ wmux    │
+ │  wmux CLI            │  :9787   │  listens 0.0.0.0   │           │         │
+ │  wmux-hook.js        │          │  npiperelay.exe ×N │ ═════════ │ \\.\pipe│
+ │  bash integration    │ ◄─────── │  (warm pool)       │ ◄──────── │  \wmux  │
+ └──────────────────────┘          └────────────────────┘           └─────────┘
+   WMUX_REMOTE=                      wmux bridge is a byte relay:
+   host.docker.internal:9787         no parsing, no auth of its own
+```
+
+Three hops, one protocol. The container speaks the same JSON-RPC (V2) and
+line (V1) protocol it would speak to a local pipe; only the socket underneath
+changes. Auth is unchanged end to end — every request still carries the wmux
+instance's pipe token, which `wmux bridge` forwards without inspecting.
+
+The bridge deliberately runs **inside WSL2**, not on Windows. From there,
+`0.0.0.0` is the WSL2 network namespace, which the container reaches through
+the Docker host gateway and the LAN does not. A Windows-side bind would expose
+the pipe to the corporate network and need a firewall rule; this needs neither.
+
+The WSL2 → Windows hop is [npiperelay.exe](https://github.com/albertony/npiperelay),
+a ~2 MB MIT-licensed binary that forwards a named pipe to its own stdin/stdout.
+WSL2 runs Windows executables via interop, so the bridge spawns it and treats
+its stdio as the socket. AF_VSOCK, a TCP listener on the Windows side and
+cross-boundary Unix sockets were all tried first: HCS-managed WSL2 VMs ignore
+`GuestCommunicationServices`, gateway IPs and firewall policy are unreliable on
+managed networks, and 9P does not forward `AF_UNIX`.
+
+## Setup
+
+### 1. Install npiperelay in the WSL2 distro
+
+Once per distro. The download is SHA-256 pinned against the release's own
+checksums file, and re-running is a no-op:
+
+```bash
+bash scripts/install-npiperelay.sh     # → ~/.local/bin/npiperelay.exe
+```
+
+The CLI also finds it on `PATH`, in `/usr/local/bin` or in `/usr/bin`, if you
+prefer to install it yourself.
+
+### 2. Start the bridge in WSL2
+
+Run this in the WSL2 distro that hosts your Docker daemon, with wmux already
+running on Windows. It stays in the foreground:
+
+```bash
+wmux bridge --wsl --port 9787
+```
+
+`--wsl` binds `0.0.0.0` instead of `127.0.0.1`. If `wmux` is not on your
+`PATH` inside WSL2, invoke the shipped CLI directly — any wmux pane exports
+`WMUX_CLI` for exactly this:
+
+```bash
+node "$WMUX_CLI" bridge --wsl --port 9787
+```
+
+### 3. Read the instance token
+
+On Windows, or in any WSL2 shell that can reach the pipe:
+
+```bash
+wmux token
+```
+
+This is the running instance's pipe token. It changes when wmux restarts.
+
+### 4. Point the container at the bridge
+
+Two variables. Add them to your `devcontainer.json`:
+
+```jsonc
+{
+  "containerEnv": {
+    "WMUX_REMOTE": "host.docker.internal:9787",
+    "WMUX_REMOTE_TOKEN": "${localEnv:WMUX_REMOTE_TOKEN}"
+  },
+  // Docker Desktop resolves host.docker.internal already; plain Docker on
+  // Linux/WSL2 does not, so map it to the gateway explicitly.
+  "runArgs": ["--add-host=host.docker.internal:host-gateway"]
+}
+```
+
+Export `WMUX_REMOTE_TOKEN` on the host before launching, so the token is not
+committed to the repo. The equivalent for a hand-rolled `docker run`:
+
+```bash
+docker run --add-host=host.docker.internal:host-gateway \
+  -e WMUX_REMOTE=host.docker.internal:9787 \
+  -e WMUX_REMOTE_TOKEN="$(wmux token)" \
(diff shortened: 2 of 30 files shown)


amirlehmam:
You asked whether to split before you did the work, so: **yes please, and I've taken two of the seven commits already** so you have less to rebase.

Cherry-picked into **1.0.0** (with `-x`, so the original SHAs are recorded):

| commit | why it went in on its own |
|---|---|
| `f5f782c` `fix(cwd)` | Standalone, additive, and 21 tests. A pwsh pane rewriting the directory a neighbouring WSL pane falls back to is a bug on its own terms — nobody needs a devcontainer to hit it. The `isPosixPath(cwd)` seed for sessions saved before the field existed is the right call; persistence types being a superset of what every writer emits is a rule this repo learned the hard way in #145. |
| `0fa84ab` `feat(config)` | Small, self-contained, and fixes a genuinely bad failure mode: the whole file is discarded on a parse error, so one stray character presents as "my setting has no effect" with nothing pointing at the file. Deduping it matters because `loadUserConfig` runs on every WSL pane spawn, and you handled that. |

Both verified on Windows, which is the side your run couldn't exercise: `npm run typecheck` clean, full suite green (1004 passing with the rest of 1.0.0 on top).

## Why the rest waits, and what it needs

Not a rejection of the feature — `wmux bridge --wsl` is a good idea and the rejected-alternatives notes (AF_VSOCK, a Windows-side listener, cross-boundary Unix sockets) are the kind of thing I'd otherwise have to rediscover. Three concrete reasons it doesn't ride into a 1.0.0:

**1. `resources/cli/wmux.js` is +1412/-445 of build output.** You state it's byte-identical to a fresh `npm run build:main`, and I believe you — but it's also 60% of the diff by volume, and it makes the PR unreviewable as a diff and unmergeable without conflicts the moment anything else touches the CLI. Which it did: #153's timeout work and my #158 changes both landed there. Please regenerate on top of current `master` rather than resolving those by hand.

**2. `c693c06` and `64fb8d8` each branch on `process.platform`, so neither side is fully exercised by a single-OS run.** Your table is honest about this and you flagged it yourself. `findNpiperelay()` splitting `$PATH` on a hard-coded `:` is a real bug on Windows and worth having regardless — but it and the `config path` / `locales path` fallback fixes belong in a small Windows-verifiable PR, not behind the bridge feature.

**3. `d1c1b58` is not the standalone commit it looks like.** "Plain copy of the `src/` file, no hand-edits" is true, but it sits *after* `9fd08ed` — so the copy carries the bridge's `raw-v1` plumbing with it. The stale-since-v0.5.2 packaged bash integration is a real bug (#149's shape: works in dev, silently absent in installs) and I want it fixed, but as a copy of a `src/` file that does not yet contain the bridge.

## What I'd take next, in order

1. **`fix(cli)` path handling** — the `path.delimiter` / `Path`-vs-`PATH` / POSIX-home-separator fixes, minus the npiperelay-specific parts if they're separable, plus `64fb8d8`'s test repair. Small, cross-platform, verifiable on Windows.
2. **`fix(shell-integration)`** — the packaged bash integration resynced from a pre-bridge `src/`. And if you're willing, the `.cmd` and `.ps1` copies you noted have drifted the same way; you're right that they're a separate change, but they're the same *class* of change and I'd rather fix all three once.
3. **`feat(devcontainer)`** — the bridge itself, on its own, with regenerated artifacts and `docs/DEVCONTAINER.md`.

Split that way I can review and ship 1 and 2 quickly, and give 3 the attention it actually deserves.

## Two things I want to flag on the bridge itself, for when it comes back

- **`wmux bridge --wsl` binding `0.0.0.0`.** Your reasoning is right that in the WSL2 network namespace this is not the LAN — but it's a line that will be read out of context by the next person to touch it, and the blast radius if that assumption ever changes (a WSL networking mode change, mirrored networking) is "wmux's control pipe is on the network". Please keep that comment as load-bearing as it is now, and if there's a way to assert the namespace rather than assume it, I'd take the extra check.
- **`wmux raw-v1`.** A CLI verb that sends arbitrary V1 lines is a wider surface than the bash integration needs. If it can be narrowed to the specific reports the integration actually emits, I'd prefer that to a general escape hatch — the V1 path predates the auth gate and I'd rather not grow it.

## One correction to your notes

> Deadlines gained a floor on the slow transports: the 5s default sat below the container→WSL2→pipe round-trip

Good find, and it's the second time that default has been wrong for a transport it wasn't designed for (#153 was the first, for browser commands). When the bridge PR comes back, I'd rather the floor be derived from the transport than added per-call site — three places choosing their own timeout is how the first one got missed.

Thanks for opening it as a draft and asking first. That saved both of us a merge conflict and got two of your fixes shipped a week earlier than they would have been.

stlindenau:
Thanks for the two cherry-picks and for the review. Split as asked, in your order. Two of them are up here; the third is in my fork for now.

- **#168** — `fix(cli)`: the config/locales path fallbacks + the bash-path test repair
- **#169** — `fix(shell-integration)`: all three packaged copies resynced
- **stlindenau/wmux#2** — the bridge, rebuilt on 1.0.0, draft in the fork

`f5f782c` and `0fa84ab` were patch-identical to your `addad6f` and `d24aed0`, so the rebase dropped both on its own. Five commits left, which is what the new branch is built from.

## Two corrections

**`findNpiperelay()` can't come out of the bridge.** You asked for PR 1 minus the npiperelay-specific parts "if they're separable" — they aren't, because the function doesn't exist on master. `git grep npiperelay upstream/master -- src/cli/wmux.ts` returns nothing; it arrives with the bridge. So the `path.delimiter` / `Path`-vs-`PATH` fix rides with #2 rather than #168. #168 is the config/locales fallbacks and the test repair only.

**The `resources/cli/wmux.js` conflict you predicted isn't there.** You expected #153's timeout work and your #158 changes to collide there. `#153` (`18940c0`) is below the PR base, and `#158` (`1dd0d9e`) touches `src/main/claude-context.ts` and friends, not the CLI — nothing in `8122b69..master` touches `src/cli/` or `resources/cli/` at all. So no conflict to resolve.

## But the artifacts are stale on master anyway

Different problem, same root as #169. On `7baf230`, a clean `npm run build:main` against the checked-in copies:

| file | drift |
|---|---|
| `resources/cli/wmux.js` | **1519 changed lines** |
| `resources/cli/wmux-hook.js` | 64 changed lines |

Last touched at `34ab219`. I've deliberately **left this alone** — #168 and #169 stay `src/`-only rather than carrying a 1519-line regenerated blob that would swamp the review of a 12-line fix. Reporting it rather than fixing it silently.

Worth saying that it doesn't currently *break* anything: `electron-builder.json` ships `dist/cli/*.js`, and the manual release process in CLAUDE.md copies from `dist/` too, so the checked-in `resources/cli/` copies are never what ships. Same for `resources/shell-integration/` in #169 — `extraResources` reads `src/shell-integration`, and `pty-manager.ts` loads `src/` in dev and `process.resourcesPath` when packaged. Neither directory is on any path a released build takes.

Which raises the real question: **who are those directories for?** As far as I can tell the only consumers are tools reading the repo checkout — mine is one of them, which is how I noticed. If that's the intent they need to be correct; if it isn't, deleting them is cleaner than syncing them. Your call, and #169 works either way.

## The CI check

This class of bug recurs because nothing enforces the copy — that's three instances now (the three shell-integration files, and both CLI artifacts). Want a CI check that `resources/` matches a fresh build? Happy to open it as a follow-up; I didn't want to bundle an unasked-for workflow change into a fix.

## On the `0.0.0.0` question

You asked about the blast radius of `--wsl` binding `0.0.0.0`, and the honest answer turned out to be "it depends on a WSL setting the code never looked at":

- **NAT mode** (the default, and what the old comment described) — the distro has its own namespace behind a Hyper-V vSwitch, so `0.0.0.0` is a private 172.x plus loopback. Reachable from containers, genuinely not from the LAN. The comment was correct and is kept.
- **Mirrored mode** (`networkingMode=mirrored`) — no namespace of its own; the distro *shares the Windows host's adapters*, LAN and VPN included. Inbound is filtered by the **Hyper-V** firewall rather than the normal Windows Firewall profile, and the recipe everyone copies for "make WSL reachable" is `Set-NetFirewallHyperVVMSetting … -DefaultInboundAction Allow`. That combination puts wmux's control pipe on the corporate network. The token still authenticates every request, so it's exposure rather than an open door — but it's a real configuration, not a hypothetical.

So #2 reads `wslinfo --networking-mode` and refuses to pick `0.0.0.0` for you under mirrored *or* unknown; `--host` still works if you want it, and warns with the mode named. The decision is a pure function with a table test, so it doesn't need WSL to run. I haven't been able to exercise the mirrored branch on real hardware — that's stated in the PR rather than glossed.

I'll close this one as superseded. Review whichever of #168/#169 you like; #2 comes up here once they've landed and it's rebased onto them.

stlindenau:
Superseded by #168, #169 and stlindenau/wmux#2 — see the comment above for the split and the two corrections.

Issue #156, opened 2026-08-12
`wmux browser` / `agent` / `pane` with no subcommand print "Unknown ... command: undefined" instead of usage


capad-xyz:
## Summary

Running a group command with no subcommand prints the literal word `undefined` instead of that command's usage. Three dispatchers do this; two others in the same file already do the right thing.

```
$ wmux browser
Unknown browser command: undefined      (exit 1)

$ wmux agent
Unknown agent command: undefined        (exit 1)

$ wmux pane
Unknown pane subcommand: undefined      (exit 1)
```

versus, in the same CLI:

```
$ wmux markdown
Usage: wmux markdown <file>  |  wmux markdown set <id> --content <text> [--title T] | --file <path>  |  wmux markdown get <id>

$ wmux config
Usage: wmux config <show|reload|path>
```

Verified on v0.46.0. Exit status is 1 in all five cases, so only the message differs.

## Cause

Each of the three interpolates `args[1]` into the error without checking that it exists, so a missing subcommand is reported as an unknown one named `undefined`:

https://github.com/amirlehmam/wmux/blob/fae21a5bbdf2edcd1c41d4c0b5372aab56caef83/src/cli/wmux.ts#L143-L147

https://github.com/amirlehmam/wmux/blob/fae21a5bbdf2edcd1c41d4c0b5372aab56caef83/src/cli/wmux.ts#L184-L188

https://github.com/amirlehmam/wmux/blob/fae21a5bbdf2edcd1c41d4c0b5372aab56caef83/src/cli/wmux.ts#L204-L206

`COMMAND_SPECS` already carries the usage string these should fall back to, and `fail()` already exists to print it — which is how `markdown` and `config` get it right. `wmux browser --help` cannot fill the gap either, since `browser` is `passthrough: true` and `--help` is treated as text to send.

## Why it is worth fixing

`wmux browser` on its own is the natural thing to type when you want to know the verbs. Getting `undefined` back reads like the CLI malfunctioned rather than like a usage error, and it is a dead end: nothing in the output points at `wmux help browser`, which is the route that does work.

Small, but it is the first thing a new user or a fresh agent hits when exploring the browser API.

## Note

`cmdBrowser` is the same function touched by #153, so if you would rather this land there than as a separate change, say the word and I will fold it in — I kept it out to avoid widening the scope of a PR you had not looked at yet.

amirlehmam:
[...]

PR #168, opened 2026-08-14
fix(cli): cross-platform config and locales paths, and a bash-path test repair

stlindenau:
First of the three splits of #166, in the order you asked for.

Three commits, `src/cli/wmux.ts` and `tests/` only. No `resources/` artifact is touched — see the note at the bottom, that turned out to be its own story.

## What is here

**`wmux config path` asks the instance.** It built the path from the caller's own home directory, which only holds when the CLI runs on the same Windows box as wmux. From WSL or a container over the bridge it printed `/home/vscode\.wmux\config.toml` — not the file wmux reads, and not a well-formed path on either OS. `loadUserConfig()` already records the real path and `config.get` returns it, so ask; the local guess stays as the no-instance answer.

**Both no-instance fallbacks are spelled for the OS in hand.** `config path`'s remaining guess and `locales path`, which has no instance-side answer to ask for at all, both hardcoded backslashes. The separator now comes from the home directory that was found rather than from `process.platform`, because the interesting case is exactly where those two disagree: a POSIX `$HOME` gets `path.posix.join`, everything else gets `path.join`. `$HOME` is now preferred over `USERPROFILE` — under WSL both are set and `USERPROFILE` is the Windows one arriving over interop.

`locales path` can still name the wrong directory across the WSL boundary. Fixing that properly needs a `locales.get` reporting the loaded directory the way `config.get` reports `path`; out of scope here.

**The bash-path helper and its test.** `bashPathCandidates()` offered `/c`, `/mnt/c`, `/cygdrive/c` in a fixed order; in a Linux checkout the bash on PATH mounts a Windows drive at `/mnt/c` if it sees one at all, so leading with `/c` put the fallback branch in charge. The order is now conditional on `process.platform`, the set is unchanged, and the probe still decides. `orchestration-status-vocab.test.ts` gains a `bashExists(hookPath)` term in its skip condition — `hasBash()` only says a bash exists, not that it can see the hook script.

`bash-path.test.ts` is in the same commit and not a separate one because it cannot survive the helper change alone: it pinned the Git-Bash-first order literally, and four assertions go red on Linux and macOS the moment the order becomes conditional. Splitting them would leave a commit that fails `npm test` for anyone not on Windows.

## What is not here, and why

You asked for this "minus the npiperelay-specific parts if they're separable". They are not: `findNpiperelay()` does not exist on master — `git grep npiperelay master -- src/cli/wmux.ts` is empty. It arrives with the bridge, so its `path.delimiter` / `Path`-vs-`PATH` fix has to ride with PR 3. Flagging rather than silently omitting.

## The `resources/cli/wmux.js` conflict you expected

It is not there. Nothing in `8122b69..master` touches `src/cli/` or `resources/cli/` — #153's timeout work (`18940c0`) is *below* the PR base, and #158 (`1dd0d9e`) is in `claude-context.ts` and friends. So this PR had no artifact conflict to resolve.

But while checking that, something else turned up, and it changes one test in this PR:

**`resources/cli/wmux.js` is 1519 diff lines behind a build of master's own `src/`** (`wmux-hook.js` is 64). It is missing `timeoutMessage` and `browserDeadline` from #153, `subcommandError` from #156, and `rename-surface` from #104. Nothing shipped is broken by this — `package.json`'s `bin`, `electron-builder.json`'s `extraResources` and step 7 of the release process all use `dist/cli/*.js`, so releases and installs get a fresh build. The stale copies are read by dev-mode Claude hooks (`claude-context.ts:394`) and by anything consuming the CLI out of a repo checkout.

`cli-config-path.test.ts` as originally written spawned `resources/cli/wmux.js`, on the stated grounds that it is "the file a released wmux actually runs" — which is the belief above, and it is wrong. It now spawns `dist/cli/wmux.js` and compiles it in `beforeAll` (~3s; `dist/` is gitignored and `npm test` runs ahead of `npm run build:main`). Building unconditionally is what makes the spawned CLI the working tree's rather than a leftover.

I have deliberately **not** regenerated the artifact here. Doing so would put 1519 lines of unrelated catch-up into a three-file fix and bury the actual change. Same class of bug as the shell-integration drift in PR 2 — happy to send a fourth PR that regenerates both `resources/cli/*.js`, or a CI check that `resources/` matches a fresh build, whichever you prefer. Tell me which and I will open it.

## Verification

- `npm run typecheck` clean (both tsconfigs).
- Full suite on this branch: **6 failed, 993 passed, 2 skipped**. Same suite on a detached `master` worktree with the same `node_modules`: **6 failed, 988 passed, 2 skipped**. The failing set is identical one-for-one — the Windows-path and live-Win32-probe cases (`shell-context-menu` ×3, `pty-ledger` ×2, `pty-manager > resolveSpawnCwd`). This PR adds 5 passing tests and takes nothing green to red.
- Linux only. `bashPathCandidates()` still has a Windows branch a Linux run cannot exercise, and `wmux config path` / `wmux locales path` want a check from a native Windows shell — that half is yours.

### DIFF
--- a/src/cli/wmux.ts
+++ b/src/cli/wmux.ts
@@ -359,7 +359,11 @@ async function cmdConfig(args: string[]): Promise<void> {
       // No reachable instance — fall through to a local guess. path.join at least
       // keeps it self-consistent with whichever filesystem we are actually on.
     }
-    console.log(path.join(os.homedir(), '.wmux', 'config.toml'));
+    const home = process.env.HOME || process.env.USERPROFILE || os.homedir();
+    const fallbackPath = home.includes('/') && !home.includes('\\')
+      ? path.posix.join(home, '.wmux', 'config.toml')
+      : path.join(home, '.wmux', 'config.toml');
+    console.log(fallbackPath);
   } else {
     console.error('Usage: wmux config <show|reload|path>'); process.exit(1);
   }
@@ -377,8 +381,14 @@ async function cmdLocales(args: string[]): Promise<void> {
   } else if (sub === 'reload') {
     print(await sendV2('config.reload'));
   } else if (sub === 'path') {
-    const home = process.env.USERPROFILE || process.env.HOME || '';
-    console.log(`${home}\\.wmux\\locales`);
+    // No locales equivalent of config.get to ask, so this stays a guess — but a
+    // guess spelled for the filesystem the CLI is on. $HOME first: under WSL both
+    // are set, and USERPROFILE is the Windows one leaking in over interop.
+    const home = process.env.HOME || process.env.USERPROFILE || '';
+    const fallbackPath = home.includes('/') && !home.includes('\\')
+      ? path.posix.join(home, '.wmux', 'locales')
+      : path.join(home, '.wmux', 'locales');
+    console.log(fallbackPath);
   } else {
     console.error('Usage: wmux locales [list|reload|path]'); process.exit(1);
   }
--- a/src/cli/wmux.ts
+++ b/src/cli/wmux.ts
@@ -342,8 +342,24 @@ async function cmdConfig(args: string[]): Promise<void> {
   } else if (sub === 'reload') {
     print(await sendV2('config.reload'));
   } else if (sub === 'path') {
-    const home = process.env.USERPROFILE || process.env.HOME || '';
-    console.log(`${home}\\.wmux\\config.toml`);
+    // Ask the instance rather than reconstructing the path here. The config lives
+    // on the Windows host, but this CLI routinely runs somewhere that cannot name
+    // it: inside WSL, or inside a devcontainer reaching wmux over the TCP bridge.
+    // There $HOME is the Linux home and the old `${home}\.wmux\config.toml` form
+    // produced `/home/vscode\.wmux\config.toml` — neither the file wmux reads nor
+    // a well-formed path on either OS. loadUserConfig() already reports the real
+    // one in `path` (user-config.ts), which config.get returns verbatim.
+    try {
+      const cfg = await sendV2('config.get');
+      if (cfg && typeof cfg.path === 'string') {
+        console.log(cfg.path);
+        return;
+      }
+    } catch {
+      // No reachable instance — fall through to a local guess. path.join at least
+      // keeps it self-consistent with whichever filesystem we are actually on.
+    }
+    console.log(path.join(os.homedir(), '.wmux', 'config.toml'));
   } else {
     console.error('Usage: wmux config <show|reload|path>'); process.exit(1);
   }
--- a/tests/helpers/bash-path.ts
+++ b/tests/helpers/bash-path.ts
@@ -20,15 +20,22 @@ export function bashPathCandidates(p: string): string[] {
   if (!drive) return [slashed];
   const letter = drive[1].toLowerCase();
   const rest = drive[2];
-  return [`/${letter}/${rest}`, `/mnt/${letter}/${rest}`, `/cygdrive/${letter}/${rest}`];
+  // Git Bash usually sees `C:` as `/c`, WSL sees `/mnt/c`, and Cygwin exposes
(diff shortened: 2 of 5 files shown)


amirlehmam:
Merged in **1.0.1**. Exactly the split I asked for, and it corrected two things I'd asserted in the #166 review.

## Where I was wrong

**The artifact conflict I predicted doesn't exist.** I told you `resources/cli/wmux.js` would conflict because "#153's timeout work and my #158 changes both landed there". You checked: `18940c0` is *below* your PR base, and #158 is in `claude-context.ts` and friends. Nothing in `8122b69..master` touches `src/cli/` or `resources/cli/`. I asserted a merge conflict without looking, which is exactly the class of thing I've been thanking other people for catching this week.

**And the belief underneath it was wrong too.** `resources/cli/wmux.js` is not "the file a released wmux actually runs" — `electron-builder.json` maps `dist/cli/wmux.js` → `cli/wmux.js`, and step 7 of the release process copies from `dist/`. You were right to re-point `cli-config-path.test.ts` at `dist/cli/wmux.js` and to build it in `beforeAll`; a test spawning the stale copy would have been testing a file nothing runs.

**`findNpiperelay()` doesn't exist on master.** My "minus the npiperelay-specific parts if they're separable" was an instruction that couldn't be followed, because the function arrives with the bridge. Flagging that instead of quietly dropping it is the right call.

## Windows verification — the half you couldn't run

Both branches, on 1.0.1 with everything else merged:

```
$ node dist/cli/wmux.js config path
C:\Users\aeont\.wmux\config.toml          # instance answered

$ WMUX_PIPE='\\.\pipe\does-not-exist' node dist/cli/wmux.js config path
C:\Users\aeont\.wmux\config.toml          # fallback, correct separators
```

And the separator heuristic across the cases that matter:

| `$HOME` | result |
|---|---|
| `/home/vscode` | `/home/vscode/.wmux/config.toml` |
| `C:\Users\aeont` | `C:\Users\aeont\.wmux\config.toml` |
| `\\server\share` | `\\server\share\.wmux\config.toml` |

The UNC row is the one I went looking for, since `home.includes('/') && !home.includes('\\')` is a heuristic rather than a parse — it lands on `path.join`, which is right.

`bashPathCandidates()`'s Windows branch is now exercised too: full suite green here at **1021 passing, 86 files**, with your 5 added.

## The decision you asked for

You offered a fourth PR — regenerate the artifacts, or a CI check. **I took both, in 1.0.1**, so you don't need to:

- `resources/cli/*.js` regenerated (it was 1745 diff lines behind by the time your branch merged).
- `npm run verify:resources` compares `resources/` against its sources, wired into CI **after `build:main`**, which is what makes `dist/cli` exist. It skips loudly rather than passing quietly when `dist/` is absent — a check that silently does nothing reads as coverage it isn't providing.
- The build-free half (`shell-integration`) is a unit test instead, so it gates a release through CI's existing `npm test` step even if the build step ever moves.

I did **not** delete the directories, which was your other option. Nothing wmux ships reads them, but you read them out of a checkout and you're unlikely to be the only one — removing a path external consumers depend on isn't something to do silently in a patch release. If you'd still prefer deletion, say so and I'll do it with a deprecation notice in a minor.

## One small thing I noticed

`cli-config-path.test.ts` compiling into `dist/` during `npm test` means the test suite mutates build output. I hit one transient collision running a build concurrently with vitest — 1008/1021 with a collection error, not reproducible in three subsequent clean runs, and impossible in CI since those steps are sequential. Not asking you to change it; the unconditional build is what makes the spawned CLI the working tree's, and that's the property worth having. Just noting it in case it ever surfaces as a mystery flake.

## Next

That leaves PR 3, the bridge itself. Whenever you're ready — the two flags from my #166 review still stand (the `0.0.0.0` namespace assumption, and narrowing `wmux raw-v1` to the reports the integration actually emits), and `resources/cli/*.js` is now clean and enforced, so a regenerated artifact in that PR will be a real diff rather than catch-up noise.

Thanks for splitting it. Both halves shipped inside a day.

Issue #137, opened 2026-07-31
still old icon on windows?


arnonuem:
In the windows task bar i still see the old icon 

<img width="126" height="64" alt="Image" src="https://github.com/user-attachments/assets/7e6db069-3099-42ea-81bc-267623261bea" />

But wmux notifications already show the new icon.

WMUX itself also shows the old icon: 

<img width="206" height="107" alt="Image" src="https://github.com/user-attachments/assets/79a0337e-c0f2-4e72-bd11-b91e0e650dd3" />

amirlehmam:
Found it, and it was a real bug — not the icon cache I initially assumed. Fixed in **v0.40.0**.

## The titlebar one (your second screenshot)

`src/renderer/assets/logo.png` was a **copy** of the app icon, and the 0.37.0 rebrand updated every shipped asset except that one. So the app was showing its own previous logo back at you, from a file that had been quietly stale for three releases. Everything else — `resources/icon.png`, `resources/icons/icon.ico`, the SVG sources — was correct, which is why notifications looked right. One file, missed.

The fix is not to re-sync the copy. The copy *was* the bug: a duplicated asset is a thing that can go stale, and it did. The titlebar now imports `resources/icons/icon-small.svg` directly, so there is one file and no second copy to drift. Incidentally that variant is the right one for the job anyway — it is drawn for ≤32px, which is exactly the size the titlebar renders at, and it drops the `>_` prompts that turn to mush down there. It also inlines as a 2.5 KB data URI, replacing a 452 KB PNG that was being squeezed from 1024px into 32 CSS pixels.

## The taskbar one (your first screenshot)

Two things, and I am confident about the first and honest about the second.

**The window icon was a single 512px PNG.** Windows asks for that icon at four sizes — 16px in Alt-Tab, 24–32px on the taskbar button, 48px in the window list, 256px in Task Manager — and one 512px representation means the shell downsamples for all of them. The `.ico` has purpose-drawn entries at each size (there is a separate *tiny* variant for 16–24px that keeps only the gradient frame and a big W, because the four-pane cross plus the W-on-a-disc turns into a grey lump at 16 device pixels). wmux now ships the `.ico` alongside the `.png` and hands Windows that instead, falling back to the PNG if it is missing.

[...]

PR #169, opened 2026-08-14
fix(shell-integration): resync all three packaged copies with src

stlindenau:
Second of the three splits of #166. One commit, three files, all of it generated from master's own `src/` — nothing is picked from the feature branch.

## The drift

`resources/shell-integration/` is a checked-in copy of `src/shell-integration/` with nothing keeping it in step. All three had drifted:

| file | behind `src/` |
|---|---|
| `wmux-bash-integration.sh` | missing the `wmux()` function |
| `wmux-cmd-integration.cmd` | missing `chcp 65001` |
| `wmux-powershell-integration.ps1` | 19 insertions, 65 deletions |

The PowerShell one is the interesting case. Its copy predates #72, so `Send-WmuxMessage` sends V1 lines with **no `auth <token> ` prefix** — and `pipe-server.ts:157` rejects every V1 command but `ping` without one. Anything running that copy reports no cwd, no git branch, no shell state and no PR; the pipe answers `unauthorized` and the integration has no way to notice. It also predates the UTF-8 console setup, the `wmux` function, deferring the PR-poll job off the first-prompt critical path, and `WMUX_STARTUP_COMMANDS` (#32).

After this, `diff -r src/shell-integration resources/shell-integration` is empty.

## Before you weigh how urgent this is

**No release path reads these files.** `electron-builder.json:50` maps `src/shell-integration` → `shell-integration`, step 7 of the release process copies `src/shell-integration/*` into staging, and `getShellIntegrationPath()` falls back to `src/shell-integration` in dev. A packaged wmux has always run the fresh source — none of the breakage above ever reached a released install.

What *does* read `resources/shell-integration/` is anything consuming wmux out of a repo checkout, which is how the drift surfaced: our devcontainer integration sources the bash copy from there directly. Same shape as `resources/cli/wmux.js`, which is 1519 diff lines behind a build of master's own `src/` for the same reason — a duplicate with no enforcement. (Details in #168.)

## The follow-up worth picking

Resyncing fixes today's drift and guarantees tomorrow's. Two ways out, happy to send either as a fourth PR — just say which:

1. **A CI step** that fails when `resources/` does not match a fresh copy/build, covering the shell scripts and `resources/cli/*.js` together.
2. **Delete `resources/shell-integration/`** and point checkout consumers at `src/shell-integration/`, since nothing wmux ships or runs reads it.

The second is the smaller repo but breaks any external consumer already on the old path, so it is your call rather than mine. (We are such a consumer and would happily move.)

## Verification

- `bash -n` clean on the packaged `.sh`; sourced in a plain bash with no error, `wmux` resolves as a function.
- Full suite on this branch: **6 failed, 988 passed, 2 skipped** — byte-identical to the same run on a detached `master` worktree with the same `node_modules`. No test reads these files, so that is expected; it confirms nothing regressed.
- The `.ps1` and `.cmd` want a native Windows check, which is yours. The specific thing to look for: with the old packaged copy the sidebar shows no cwd or branch in a PowerShell pane launched from a repo checkout, and with this one it should.
- A manual run of `npm run test` in git bash on windows showed no failing tests (Claude has been running in a devcontainer on WSL to do the heavy lifting of this PR)

### DIFF
--- a/resources/shell-integration/wmux-bash-integration.sh
+++ b/resources/shell-integration/wmux-bash-integration.sh
@@ -4,6 +4,10 @@
 
 export WMUX=1
 
+# wmux CLI shortcut — Claude Code and users can just type: wmux browser open <url>
+wmux() { node "$WMUX_CLI" "$@"; }
+export -f wmux
+
 _wmux_report() {
     local msg="$1"
     # Write to temp file for main process to pick up
--- a/resources/shell-integration/wmux-cmd-integration.cmd
+++ b/resources/shell-integration/wmux-cmd-integration.cmd
@@ -5,6 +5,10 @@ REM Reports CWD via OSC 9 escape sequence embedded in prompt
 REM Set WMUX env var
 set WMUX=1
 
+REM UTF-8 code page so multi-byte input (Korean, Japanese, Chinese, emoji)
+REM round-trips through conpty correctly.
+chcp 65001 >nul 2>&1
+
 REM Set prompt to include OSC 9 with current directory
 REM ESC]9;9;PATH ESC\ then normal prompt
 prompt $e]9;9;$P$e\$P$G
--- a/resources/shell-integration/wmux-powershell-integration.ps1
+++ b/resources/shell-integration/wmux-powershell-integration.ps1
@@ -3,10 +3,25 @@
 
 $env:WMUX = "1"
 
-# Named pipe client helper
+# UTF-8 I/O so multi-byte input (Korean, Japanese, Chinese, emoji, accents)
+# survives the conpty round-trip cleanly.
+try {
+    [Console]::InputEncoding = [System.Text.UTF8Encoding]::new()
+    [Console]::OutputEncoding = [System.Text.UTF8Encoding]::new()
+    $OutputEncoding = [System.Text.UTF8Encoding]::new()
+    $PSDefaultParameterValues['Out-File:Encoding'] = 'utf8'
+} catch {}
+
+# wmux CLI shortcut — Claude Code and users can just type: wmux browser open <url>
+function wmux { node "$env:WMUX_CLI" @args }
+
+# Named pipe client helper. State updates carry an "auth <token> " prefix —
+# wmux injects WMUX_PIPE_TOKEN into every shell it spawns, and the pipe server
+# rejects unauthenticated V1 commands (issue #72).
 function Send-WmuxMessage {
     param([string]$Message)
     try {
+        if ($env:WMUX_PIPE_TOKEN) { $Message = "auth $($env:WMUX_PIPE_TOKEN) $Message" }
         $pipe = New-Object System.IO.Pipes.NamedPipeClientStream(".", "wmux", [System.IO.Pipes.PipeDirection]::InOut)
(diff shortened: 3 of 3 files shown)


amirlehmam:
Merged in **1.0.1**.

The PowerShell one is worse than "drifted", and I want it recorded properly: that copy predates #72, so `Send-WmuxMessage` sent V1 lines with **no `auth <token> ` prefix**, and `pipe-server.ts` rejects every V1 command but `ping` without one. So anything running it reported no cwd, no git branch, no shell state and no PR — and the failure is completely silent in both directions. The pipe writes `unauthorized` back into a socket the integration never reads, and the integration's own error handling is `catch { }` with the comment "Silently ignore pipe errors". Two independent reasons nobody would ever see a message.

Confirmed on the merged file:

```powershell
if ($env:WMUX_PIPE_TOKEN) { $Message = "auth $($env:WMUX_PIPE_TOKEN) $Message" }   # line 24
if ($pipeToken) { $msg = "auth $pipeToken $msg" }                                  # line 123
```

Both call sites, and `diff -r src/shell-integration resources/shell-integration` is empty on Windows.

You were right to lead with "no release path reads these files" rather than let me over-rate the severity. `electron-builder.json:50` maps `src/shell-integration`, step 7 of the release process copies from `src/`, and `getShellIntegrationPath()`'s dev fallback resolves to `src/` too. No released install ever ran the broken copy. It's a checkout-consumer bug, which is a smaller blast radius and a real one.

## The follow-up, decided

You offered a CI step or deletion. **I took the CI step, in this release**, so you don't need to send it:

- `tests/unit/resources-sync.test.ts` — byte-compares the directories, newline-normalised. No build needed, so it runs on every `npm test`, which is what gates a release in CI.
- `npm run verify:resources` — the same check plus `resources/cli/*.js` against `dist/cli/*.js`, wired into CI after `build:main`.

One test in there names the auth prefix directly rather than leaving it to the byte comparison, because that's the property whose absence was invisible. A future edit that drops it should fail as "the PowerShell integration no longer authenticates", not as "the files differ".

**Not deletion**, though it's the tidier repo. Nothing wmux ships reads either directory, but you read them out of a checkout and you're unlikely to be the only one; removing a path external consumers depend on isn't a thing to do silently in a patch release. If you'd still rather move, say so — I'll deprecate the path in a minor and drop it in the one after, and you get to be the reason it's safe.

Worth saying plainly: this is #137 again. That was a second copy of the app icon going stale for three releases, and the lesson written down at the time was that a duplicate has exactly one source. I resynced that one and didn't add a check, so the same failure just recurred in two other directories. It's checked now.

## What I verified, and what I didn't

Verified: `diff -r` empty, `bash -n` clean, both auth call sites present, full suite green at **1021 passing / 86 files**, and `verify:resources` fails as intended when I deliberately perturb a file.

Not verified: your specific end-to-end check — a PowerShell pane launched from a repo checkout showing cwd and branch in the sidebar. The packaged copy is now byte-identical to `src/`, and `src/` is what every released install has always run and does work, so the inference is short. But it is an inference, and you should treat it as one if you're relying on it.

Thanks — the auth-token finding is the kind of thing that stays broken for years because nothing ever raises its hand.

Issue #19, opened 2026-06-02
Font Family - Font Size settings is not persist


HellCatVN:
Font Family - Font Size settings is not persist after install new version or receive new update version but not update yet 

<img width="730" height="512" alt="Image" src="https://github.com/user-attachments/assets/192119e3-1d70-4e6f-a72b-8320cb289465" />

<img width="928" height="255" alt="Image" src="https://github.com/user-attachments/assets/35935f67-c2c9-49d8-bb3f-591b75d53799" />

amirlehmam:
Fixed in **v0.8.1** 🎉

Root cause: settings (font family/size, theme, shortcuts) were stored in the renderer's `localStorage`, which the browser engine scopes to the page origin. Since wmux ships as a portable zip you extract to a new folder for each version, that origin changed on every update and the saved settings got orphaned — so your customizations looked like they reset.

Settings now persist to `%APPDATA%\wmux\settings.json`, which is independent of the install folder and survives updates. Your existing settings are migrated automatically the first time you launch 0.8.1.

Grab it here: https://github.com/amirlehmam/wmux/releases/tag/v0.8.1 (remember to right-click → Unblock the zip after download). Thanks for the clear report!

Issue #81, opened 2026-07-10
Running state misleading


arnonuem:
When i run my coding tool by executing its binary the state of the session immediately goes into running.

But that coding tool is idling. Would be nice to make that status more fine grained somehow otherwise its not that helpful.

I have multiple sessions all with that coding tool but not all coding tools are really doing anything at the same time they are idling. But the tab shows running. I guess you determine it by whether a process is started or not.

I know its hard to find out - but maybe other tools can kind of hook into the state model by providing information like flag files or whatever. 

Would love to have some sort of a solution here.

arnonuem:
Whats also interesting is, that when the app started to fire up some shell commands for example for merging pullrequests and stuff - the session went to idle o_O

arnonuem:
As a workaround it would be helpful if i could set the state manually RUNNING or IDLE - its not ideal but at least it would reflect the truth then. RightClick toggle state or similar.

tawman:
@arnonuem have you installed the latest release with [PR 79](https://github.com/amirlehmam/wmux/pull/79) merged? it had fixes on the status being wrong

amirlehmam:
Your workaround suggestion is exactly what shipped in [v0.21.0](https://github.com/amirlehmam/wmux/releases/tag/v0.21.0): **right-click a workspace → Status Indicator → *Auto (detected)* / *Pin as Running* / *Pin as Idle***. A pinned status wins over all auto-detection, and *Auto* switches back to the heuristics anytime. So when your coding tool idles while the shell still technically has a foreground process, you can pin the truth by hand.

On the detection side, @tawman's point stands: make sure you're on **v0.20.0+** — [PR #79](https://github.com/amirlehmam/wmux/pull/79) reworked workspace status (coordinator-owned + self-healing), and v0.21.0 also clears a stuck "Running" badge when the PTY process actually exits. That should fix both the "immediately shows running" and the "went idle while firing shell commands" cases you saw.

Longer term, your idea of letting tools hook into the state model is interesting — wmux already exposes a CLI (`wmux set-status`, `wmux hook --event`) that agents use to report activity, so a flag-file/CLI-based state protocol for arbitrary tools is a natural extension. If the manual pin + v0.21.0 detection still doesn't reflect reality for your setup, tell me which tool you're running and I'll look at wiring it up properly.

Closing as shipped — comment and I'll reopen if it still misbehaves after updating.

amirlehmam:
[...]

PR #170, opened 2026-08-14
feat(devcontainer): drive wmux from a container over the TCP bridge

stlindenau:
Third and last of the three PRs #166 was split into. #168 and #169 are merged, so this is the remainder the split was for: the bridge itself, on its own, with regenerated artifacts and `docs/DEVCONTAINER.md`.

Rebuilt on current master (`3078b48`, 1.0.1) rather than rebased — four of the eleven commits on the old branch are now upstream and had nothing left to carry:

| old commit | why it is gone |
|---|---|
| `fix(cli): ask the instance for the config path` | merged as `1bc0089` |
| `fix(cli): improve path handling…` | merged as `7a90462`, minus the npiperelay half, which folds into commit 1 here — it is a fix to a function this branch introduces |
| `test(bash-path): assert the candidate order…` | merged as `1283433`; the three test files are byte-identical to what is on master |
| `fix(shell-integration): sync the packaged bash integration` | superseded by `c573842`, which did all three copies |

`tests/unit/cli-config-path.test.ts` is master's version untouched — spawning `dist/cli/` with a `beforeAll` build is better than what #166 had, and nothing here needs to touch it.

## What it does

`wmux bridge` exposes the local pipe on TCP so a devcontainer can drive the wmux running on the Windows host (issue #19, on top of the bridge from #78). The container has no `\\.\pipe\wmux`, so every CLI call and every Claude Code hook failed silently — the sidebar sat on "Running" while the agent idled.

Rejected alternatives are written up in `docs/DEVCONTAINER.md` (AF_VSOCK, a Windows-side listener, cross-boundary Unix sockets).

## Changes since #166

Four, from the review there.

### 1. The `0.0.0.0` bind is now checked, not assumed

#166 asked what the blast radius of `--wsl` binding `0.0.0.0` was. The honest answer is that it depends on a WSL setting the code never looked at:

- **NAT mode** (WSL2's default, and what the old comment described). The distro has its own network namespace behind a Hyper-V vSwitch. `0.0.0.0` is `eth0` on a private 172.x plus loopback — reachable from containers via the host gateway, genuinely not reachable from the LAN, no firewall rule involved. The comment was right, and it is kept verbatim as the NAT explanation.
- **Mirrored mode** (`networkingMode=mirrored` in `.wslconfig`). The distro has no namespace of its own — it *shares the Windows host's interfaces*, including the physical LAN adapter and any VPN adapter. `0.0.0.0` there is a bind on the corporate network. Inbound traffic to a mirrored distro is filtered by the **Hyper-V firewall**, not the ordinary Windows Firewall profile, and the widely-copied "make WSL reachable" recipe is `Set-NetFirewallHyperVVMSetting -Name '{40E0AC32-46A5-438A-A0B2-2B479E8F2E90}' -DefaultInboundAction Allow`. Mirrored **plus** that setting puts wmux's control pipe on the LAN. The pipe token still authenticates every request, so it is exposure rather than an open door — but it is a configuration people actually run, not a hypothetical.

So the mode is read from `wslinfo --networking-mode` (WSL 2.0.5+) and the bind follows from it:

| mode | `--wsl` does |
|---|---|
| `nat` | binds `0.0.0.0`, and says why that is safe here |
| `mirrored` | **refuses**, explains that the distro shares the host's adapters, points at `--host <addr>` |
| unknown / no `wslinfo` | **refuses** — treated as unknown, not as NAT |
| `--host` given explicitly | honoured; a non-loopback bind warns, and the warning names the mode |

The decision is a pure function (`src/cli/wsl-network.ts`), table-tested over mode × in-WSL2 × explicit-host, so none of it needs WSL to run.

One thing worth flagging, because the obvious spelling of the WSL2 probe is wrong in a way that ships easily: **a Linux container on a Windows host runs on the WSL2 kernel.** Its `/proc/sys/kernel/osrelease` reads `microsoft-standard-WSL2` while it is not a WSL2 distro and has no interop to reach a Windows host with. Matching osrelease alone would bind `0.0.0.0` in every devcontainer on Windows — the exact opposite of the intent. `isWsl2()` requires the interop vars too, and there is an end-to-end test for it that passes precisely because CI-style containers are that case.

### 2. `raw-v1` is an allowlist

It was a generic passthrough: whatever line you gave it reached the V1 handler. That is a standing side door — every V1 command added later becomes container-reachable the day it lands, without anyone deciding it should be. Now restricted to the six verbs the shell integration actually emits (`report_pwd`, `report_git_branch`, `clear_git_branch`, `report_shell_state`, `ports_kick`, `report_startup_command`); anything else exits non-zero naming the accepted set.

The test derives the expected verbs by scanning `_wmux_report` call sites in `src/shell-integration/wmux-bash-integration.sh`, so the allowlist and the integration cannot drift apart silently.

### 3. One transport-derived deadline, not two

`wmux.ts` derived its deadline from `remoteTarget || usesNpiperelay()`; `wmux-hook.ts` had its own `remote ? 30000 : 5000`. Same intent, two spellings — and the hook's did not know about npiperelay, so a hook firing from a WSL shell armed the 5s local-pipe timer for a hop that measures ~7s and killed a call that was going to succeed.

Both now describe their connection the same way and ask `transport-deadline.ts`. No behaviour change for a local pipe: the floor only ever raises a deadline.

### 4. Artifacts regenerated

`resources/cli/*.js` from a clean `npm run build:main` on the final tree, as the last commit.

`42e3cca` changed what that commit means, for the better. It is now an invariant `npm run verify:resources` checks byte-for-byte rather than bookkeeping that can slip, and — more to the point of the split — it is finally *readable*. The same regeneration in #166 was +1412/-445, because the checked-in copy had drifted ~1700 diff lines behind its own source and the bridge's output was buried in unrelated rot. Against today's master it is +443/-36 plus the two new modules, and all of it is the bridge.

That check also covers `resources/shell-integration/`, which is why the bash integration's `WMUX_REMOTE` branch is applied to both copies in commit 1: #169 resynced from a pre-bridge `src/`, so the packaged copy has the `wmux` shim but not the transport, and leaving it that way fails `npm test`.

## One thing this turned up

Splitting those two modules out exposed a packaging gap. `extraResources` ships `dist/cli/wmux.js` and `dist/cli/wmux-hook.js` **file-by-file**, not the directory — harmless while each was self-contained, but a shared sibling module is `MODULE_NOT_FOUND` on the first line of `wmux ping` and of every Claude Code hook in an installed build. `npm run dev` cannot show it; the module is sitting right there in `dist/`.

That is a worse failure than the one `packaging.test.ts` already guards (a missing *resource* is a silently-absent feature behind an `existsSync` warning), so the check is derived the same way that file's `process.resourcesPath` scan is: walk the relative imports out of each shipped CLI entry point and require each to be packaged too. `wsl-network.js` was already missing when I added it — found by the check, not by review.

## Verification

`npm run typecheck` clean. `npm run lint` reports 29 problems (16 errors, 13 warnings) — identical to master, none in files this branch touches. `npm run verify:resources` passes.

Full suite against a detached `upstream/master` worktree sharing the same `node_modules`:

| | test files | tests |
|---|---|---|
| `master` (`3078b48`) | 4 failed, 82 passed | 6 failed, 1005 passed, 2 skipped |
| this branch | 4 failed, 87 passed | 6 failed, **1045 passed**, 6 skipped |

The failing set is **identical one-for-one** — the Windows-path and live-`Win32_Process` cases (`orchestration-watcher`, `pty-ledger` ×2, `pty-manager`, `shell-context-menu` ×3) that do not pass on Linux on master either. The 4 added skips are the npiperelay round-trip cases, which need a real WSL2 distro; the suites they live in do run, at 65 passed / 4 skipped across the eight bridge-related files.

Checked by hand from the container this was built in:

- `node resources/cli/wmux.js --help` — the packaged copies resolve their new siblings
- `raw-v1 definitely_not_a_verb …` → exits 1, naming the six accepted verbs
- `bridge --wsl` → exits 1, naming what it looked for and pointing at `--host`. This container *is* the false-positive case from §1: WSL2 kernel in `osrelease`, no interop vars, correctly refused
- `resources/cli/*.js` byte-identical to a fresh `npm run build:main`

**Not verified, and I am not going to imply otherwise: the mirrored-mode branch.** This was developed on a NAT host. Exercising it needs `networkingMode=mirrored` in `.wslconfig` and a `wsl --shutdown` on a scratch machine, where `wslinfo --networking-mode` should print `mirrored`, `--wsl` should refuse, `--host 0.0.0.0` should bind with the upgraded warning, and `ip addr` inside the distro should show the host's LAN adapter — which is the observation that makes the risk concrete rather than argued. The refusal is the safe default whether or not anyone gets to that.

Draft until you have had a look at the split — happy to reshape the commit series.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -215,6 +215,8 @@ rm -rf ../wmux-release-staging/resources/themes && cp -r resources/themes ../wmu
 rm -rf ../wmux-release-staging/resources/sounds && cp -r resources/sounds ../wmux-release-staging/resources/sounds
 mkdir -p ../wmux-release-staging/resources/cli && cp dist/cli/wmux.js ../wmux-release-staging/resources/cli/wmux.js
 cp dist/cli/wmux-hook.js ../wmux-release-staging/resources/cli/wmux-hook.js   # Claude hooks exec this via bare node — MUST ship outside the asar (missing until 0.29.1 → sidebar stuck on "Running", issue #81)
+cp dist/cli/transport-deadline.js ../wmux-release-staging/resources/cli/transport-deadline.js   # required by BOTH files above; omitting it is MODULE_NOT_FOUND on the first line, not a degraded feature
+cp dist/cli/wsl-network.js ../wmux-release-staging/resources/cli/wsl-network.js                 # required by wmux.js (bridge bind selection)
 rm -rf ../wmux-release-staging/resources/shell-integration && mkdir -p ../wmux-release-staging/resources/shell-integration
 cp -r src/shell-integration/* ../wmux-release-staging/resources/shell-integration/
 rm -rf ../wmux-release-staging/resources/wmux-orchestrator && cp -r resources/wmux-orchestrator ../wmux-release-staging/resources/wmux-orchestrator
@@ -296,6 +298,7 @@ rm -rf .asar-staging build-out /tmp/asar-verify ../wmux-release-staging
 - [ ] wmux-orchestrator plugin copied to release staging
 - [ ] cli-bin + cli-bin-ps copied to release staging (issue #154)
 - [ ] opencode-plugin copied to release staging (issue #149) — `npm test` now derives this from `process.resourcesPath` usage in `src/main/`, so a new runtime resource that isn't in `extraResources` fails the release build
+- [ ] `resources/cli/` holds all four files — `wmux.js`, `wmux-hook.js`, `transport-deadline.js`, `wsl-network.js`. The CLI is packaged file-by-file, so a shared module has to be listed on its own; `npm test` derives this from the relative imports in `src/cli/`
 - [ ] rcedit applied (icon + version metadata) — `{ rcedit }` destructured
 - [ ] `latest.yml` generated (sha512 + size of the final zip) and uploaded as a release asset — electron-updater 404s without it (issue #68)
 - [ ] Zip created and uploaded to GitHub release
--- a/README.md
+++ b/README.md
@@ -326,6 +326,12 @@ wmux --remote 127.0.0.1:9787 --token <TOKEN> list-workspaces
 wmux --remote 127.0.0.1:9787 --token <TOKEN> new-workspace --title "api"
 # Or set once: WMUX_REMOTE=127.0.0.1:9787 and WMUX_REMOTE_TOKEN=<TOKEN>
 
+# From a devcontainer — same transport, no SSH tunnel. Run the bridge in WSL2
+# (--wsl binds 0.0.0.0 there, reachable from the container, not from the LAN):
+wmux bridge --wsl                  # relays to \\.\pipe\wmux via npiperelay.exe
+# Then in the container: WMUX_REMOTE=host.docker.internal:9787 + WMUX_REMOTE_TOKEN
+# Full setup: docs/DEVCONTAINER.md
+
 # Agents
 wmux agent spawn --cmd "claude --resume abc" --label "Research"
 wmux agent spawn-batch --json '[{"cmd":"claude","label":"Agent 1"},{"cmd":"claude","label":"Agent 2"}]'
@@ -345,9 +351,14 @@ Connect to `\\.\pipe\wmux` for programmatic control. Two protocols supported:
 report_pwd <surface_id> <path>
 report_git_branch <surface_id> <branch> [dirty]
 report_shell_state <surface_id> idle|running|interrupted
+report_startup_command <surface_id> <command>   # how to restore this surface; must be cwd-independent
 notify <surface_id> <text>
 ping
 ```
+
+`wmux raw-v1 "<line>"` sends any of these through the CLI's transport, which is
+what lets a shell with no reachable pipe — inside a devcontainer — still report.
+See [docs/DEVCONTAINER.md](docs/DEVCONTAINER.md).
 
 **V2** (JSON-RPC, used by CLI and automation):
 ```json
--- a/docs/DEVCONTAINER.md
+++ b/docs/DEVCONTAINER.md
@@ -29,10 +29,17 @@ line (V1) protocol it would speak to a local pipe; only the socket underneath
 changes. Auth is unchanged end to end — every request still carries the wmux
 instance's pipe token, which `wmux bridge` forwards without inspecting.
 
-The bridge deliberately runs **inside WSL2**, not on Windows. From there,
-`0.0.0.0` is the WSL2 network namespace, which the container reaches through
-the Docker host gateway and the LAN does not. A Windows-side bind would expose
-the pipe to the corporate network and need a firewall rule; this needs neither.
+The bridge deliberately runs **inside WSL2**, not on Windows. Under WSL2's
+default **NAT** networking, `0.0.0.0` there is the distro's own network
+namespace, which the container reaches through the Docker host gateway and the
+LAN does not. A Windows-side bind would expose the pipe to the corporate network
+and need a firewall rule; this needs neither.
+
+That holds for NAT and not for **mirrored** networking, where the distro shares
+the Windows host's interfaces rather than having a namespace of its own — see
+[Mirrored networking](#mirrored-networking) below. `wmux bridge --wsl` reads the
+mode with `wslinfo --networking-mode` and only picks `0.0.0.0` once it has
+confirmed NAT.
 
 The WSL2 → Windows hop is [npiperelay.exe](https://github.com/albertony/npiperelay),
 a ~2 MB MIT-licensed binary that forwards a named pipe to its own stdin/stdout.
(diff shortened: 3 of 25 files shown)


amirlehmam:
Merged in **1.1.0**. All seven commits, as submitted.

This is the third of three, and the split was worth it exactly as you predicted: against today's master the artifact commit is **+443/−36** and every line of it is the bridge. The same regeneration in #166 was +1412/−445 of accumulated rot with the actual change buried inside. I could review this one.

## The four review points

**1. The `0.0.0.0` bind.** I asked what the blast radius was and said the comment should stay load-bearing. You did better than that — you went and found that the comment was only true under one of two networking modes, and that the other one is a configuration people actually run.

The mirrored-mode detail is the part I did not know: the distro has no namespace of its own, so `0.0.0.0` is the host's real interfaces including LAN and VPN, and inbound is gated by the **Hyper-V** firewall rather than the ordinary profile — which is why the widely-copied `Set-NetFirewallHyperVVMSetting … -DefaultInboundAction Allow` recipe is the thing that turns exposure into reachability. That is a much sharper answer than the one I asked for.

Treating `unknown` as unknown rather than folding it into the NAT default is the right call and the one that's easy to get wrong under pressure to make `--wsl` "just work" on WSL 1.x. And honouring an explicit `--host 0.0.0.0` under mirrored mode is right too: the user named the address, so describing the choice beats overruling it.

**The `isWsl2()` false positive is the best catch in the PR.** A Linux container on a Windows host runs on the WSL2 kernel, so `/proc/sys/kernel/osrelease` says `microsoft-standard-WSL2` while it is not a distro and has no interop to reach a host with. Matching osrelease alone — the obvious spelling — would bind `0.0.0.0` in *every devcontainer on Windows*, which is the precise inverse of the feature's intent. Requiring the interop vars too, and having the end-to-end test pass *because* CI containers are that case, is the kind of thing that only gets found by someone actually running it.

**2. `raw-v1` as an allowlist.** Yes. My concern was the standing side door, and you named the property better than I did: every V1 command added later becomes container-reachable the day it lands, without anyone deciding it should be. Deriving the expected verbs by scanning `_wmux_report` call sites in the integration means the two cannot drift apart silently — same shape as the `verify:resources` invariant, and the right instinct.

**3. One transport-derived deadline.** This is what I meant by "derived from the transport rather than added per call site", and the bug it fixes is worse than the duplication: the hook's copy didn't know about npiperelay, so a hook firing from a WSL shell armed a 5s timer for a ~7s hop and killed calls that were going to succeed. That's the third time a deadline default has been wrong for a transport it wasn't designed for (#153 was the first). A floor that only ever raises is the right shape.

**4. Artifacts.** Regenerated, and `verify:resources` passes.

## The packaging gap

This is the most valuable thing in the PR and it isn't the feature.

`extraResources` ships the CLI entry points **file-by-file**, not the directory. Harmless while each was self-contained; the moment there's a shared sibling it's `MODULE_NOT_FOUND` on the first line of `wmux ping` and of every Claude Code hook — in installed builds only, invisible to `npm run dev` because the module is sitting right there in `dist/`.

I reproduced both directions before merging, because a packaging claim deserves it:

```
# copy ONLY what extraResources declares, run from that dir
$ node <sim>/cli/wmux.js ping
pong                                   # all four modules resolve

# drop one sibling
$ rm <sim>/cli/transport-deadline.js && node <sim>/cli/wmux.js ping
node:internal/modules/cjs/loader:1424
  throw err;                           # exactly the failure you describe
```

Same class as #149 and #81 — works in dev, silently absent when installed — and it has now bitten this project three times. Deriving the check from the imports rather than a hardcoded list is what makes it hold, and the self-check test ("finds the sibling imports it is meant to be checking") is the detail that keeps it from quietly matching nothing. That it caught `wsl-network.js` before review did is the whole argument.

## Windows verification — your half

Native Windows 11, Core Ultra 9 275HX / RTX 5090, on the merge commit:

| check | result |
|---|---|
| `npm run typecheck` | clean |
| `npm test` | **1059 passed, 6 skipped** (the npiperelay cases) |
| `npm run verify:resources` | passes |
| `bridge --wsl` on native Windows | refuses, exit **1**, names what it looked for |
| `raw-v1 definitely_not_a_verb` | refuses, exit **1**, names the six verbs |
| `wmux ping` | `pong` — no regression from the module split |
| packaged-layout simulation | resolves siblings (above) |

So the two `process.platform` branches you flagged as unexercised are now exercised.

## Mirrored mode

Noted, and I'm not going to pretend otherwise either — it stays unverified for now. I don't have a mirrored host and I'm not reconfiguring `.wslconfig` on this machine to find out.

You're right that the refusal is the safe default whether or not anyone gets there: the untested branch's failure mode is "refuses to start a bridge that would have worked", which costs a confused user an error message, versus "binds the control pipe to the LAN", which doesn't. That asymmetry is what makes shipping it acceptable.

If anyone reading this runs `networkingMode=mirrored`, the four observations worth reporting are in the PR body and I'd take them as an issue.

## Two small things, not blocking

- `report_startup_command` stores per surface, and `startupCommands` already round-trips through `surface-slice`'s save/restore — so a restored pane re-enters its container across a full session restore, not just a remount. Worth stating because #145 was exactly the case where a writer emitted a field persistence didn't know about.
- The `--cap`/RAM tuning story for the bridge path isn't documented in `DEVCONTAINER.md`, and the round-trip cost makes cache temperature matter more over the bridge than locally. Not a blocker, and possibly better as a `tuning.md` cross-reference than anything here.

---

That's #166 fully landed, in three reviewable pieces, with two bugs found along the way that had nothing to do with the feature. Thanks for taking the split — it was the right call and you did the harder half of it.

PR #171, opened 2026-08-14
fix(cli): match raw-v1 on the verb, not on the whole argv entry

stlindenau:
## What breaks

Every V1 report the bash shell integration sends from a devcontainer is dropped by the CLI before it reaches the wire. In the sidebar that shows up as a pane with no cwd, no git branch, no shell state and no restore command. All six allowlisted verbs are affected.

## Why

`src/shell-integration/wmux-bash-integration.sh` builds the V1 line as one string and passes it quoted:

```bash
( wmux raw-v1 "report_pwd $surface_id $(pwd)" >/dev/null 2>&1 & )
```

so `args[1]` is the whole line. `cmdRawV1` checked `args[1]` against `RAW_V1_VERBS`, which holds bare verbs:

```ts
const problem = rawV1Error(args[1]);   // "report_pwd surf-1 /home/me"
```

`RAW_V1_VERBS.includes("report_pwd surf-1 /home/me")` is false, so the CLI printed an error and `process.exit(1)` before `sendV1` was reached.

Nothing surfaced it. The caller fires into `>/dev/null 2>&1 &`, so the error went nowhere, and `ping` / `identify` / `list-surfaces` stayed healthy the whole time because the transport was never the problem.

Provenance: the allowlist (`e32a6dd`) landed *after* the quoted single-argument call site (`3d6d52d`), and `tests/unit/raw-v1-allowlist.test.ts` only ever called `rawV1Error()` with bare verbs — a shape no real caller uses. Its comment even asserted "args[1] is the verb; the surface id and payload follow as separate argv entries", describing a caller that does not exist. That is why it shipped green.

## The fix

Take the first whitespace token of the joined line:

```ts
export function rawV1Parse(args: string[]): { line: string; verb: string } {
  const line = args.slice(1).join(' ');
  return { line, verb: line.trim().split(/\s+/)[0] ?? '' };
}
```

This is not a loosening of the allowlist. `pipe-server.ts` `handleV1()` already parses the command as the first token, so the allowlist now agrees with the parser it guards rather than with one caller's argv habits. Concretely:

- a pre-split argv (`wmux raw-v1 report_pwd surf-1 /tmp`) is unchanged — both shapes produce the same line and the same verb
- `"report_pwd notify"` is a `report_pwd` whose surface id is `notify`, not a smuggled `notify`
- an unknown first token is still refused, now naming just the verb instead of echoing the whole line back

`rawV1Error` is untouched, so every existing assertion still holds.

## Tests

The gap *is* the bug, so the new cases target the argv layer specifically: the one-argument shape, the equivalence of both shapes, a path containing spaces, a non-allowlisted verb in the one-argument shape, and every verb scraped out of the shipped shell script — sent the way the script sends them, not the way the old test pretended.

Six of the fifteen fail without the source change; all fifteen pass with it. Verified in both directions.

## Artifacts

`resources/cli/wmux.js` is regenerated (`npm run build:main`, copy, `npm run verify:resources`). It is what actually executes, so a source-only fix would ship nothing.

## Verification

`npm run typecheck` clean, `npm run verify:resources` clean. Against a dead port, so nothing reaches a real server:

```
# before — rejected locally, never sent
$ WMUX_REMOTE=127.0.0.1:1 node <old>/wmux.js raw-v1 "report_git_branch surf-1 main"
raw-v1: report_git_branch surf-1 main is not a passthrough command...

# after — allowlist passes, transport is reached
$ WMUX_REMOTE=127.0.0.1:1 node <new>/wmux.js raw-v1 "report_git_branch surf-1 main"
wmux is not running

# still refused
$ ... raw-v1 "notify surf-1 hello"
raw-v1: notify is not a passthrough command...
```

Full suite on a Windows host: **1069 passed, 0 failures.**

Earlier runs of this branch showed 6 failures (`shell-context-menu`, `pty-ledger`, `pty-manager`). Those were run from a Linux devcontainer, where the ConPTY and Windows-shell suites cannot execute; they were confirmed identical on pristine `master` at the time. On Windows they pass. Nothing in this change touches them either way — the earlier failures were the environment, not the branch.

### DIFF
--- a/resources/cli/wmux.js
+++ b/resources/cli/wmux.js
@@ -9,6 +9,7 @@ exports.timeoutMessage = timeoutMessage;
 exports.browserRequest = browserRequest;
 exports.subcommandError = subcommandError;
 exports.rawV1Error = rawV1Error;
+exports.rawV1Parse = rawV1Parse;
 const net_1 = __importDefault(require("net"));
 const fs_1 = __importDefault(require("fs"));
 const os_1 = __importDefault(require("os"));
@@ -996,13 +997,38 @@ function rawV1Error(verb) {
         return null;
     return `raw-v1: ${verb} is not a passthrough command. Accepted: ${exports.RAW_V1_VERBS.join(', ')}`;
 }
+/**
+ * Split a `raw-v1` argv into the V1 line to send and the verb to check.
+ *
+ * The two callers disagree about argv, and both are legitimate. A hand-typed
+ * call splits naturally — `wmux raw-v1 report_pwd surf-1 /tmp` — but
+ * wmux-bash-integration.sh builds the line as a single string and passes it
+ * quoted: `wmux raw-v1 "report_pwd $surface_id $(pwd)"`. There the whole line
+ * is args[1].
+ *
+ * Checking args[1] against the allowlist therefore rejected every report the
+ * shell integration ever sent: "report_pwd surf-1 /tmp" is not in RAW_V1_VERBS,
+ * so the CLI exited 1 before sendV1 was reached. Nothing reached the wire, and
+ * because the integration fires into `>/dev/null 2>&1 &` there was no symptom
+ * beyond a sidebar that never showed a cwd or a branch.
+ *
+ * Taking the first whitespace token is not a loosening — it is what the server
+ * already does. pipe-server.ts handleV1() parses the command as the first token
+ * of the line, so this makes the allowlist agree with the parser it guards
+ * rather than with one caller's argv habits.
+ */
+function rawV1Parse(args) {
+    const line = args.slice(1).join(' ');
+    return { line, verb: line.trim().split(/\s+/)[0] ?? '' };
+}
 async function cmdRawV1(args) {
-    const problem = rawV1Error(args[1]);
+    const { line, verb } = rawV1Parse(args);
+    const problem = rawV1Error(verb);
     if (problem) {
         console.error(problem);
         process.exit(1);
     }
-    console.log(await sendV1(args.slice(1).join(' ')));
+    console.log(await sendV1(line));
 }
 // ─── Declared agent state (issue #128) ───────────────────────────────────────
 // The reporting side of the protocol. An agent running inside a wmux pane can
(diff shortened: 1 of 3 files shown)


amirlehmam:
Merged, shipped in **1.2.0**. Thank you — this is a model bug report attached to a fix.

## The part I want on the record

You didn't just fix it, you explained why it shipped green, and that explanation is the more valuable half:

> Its comment even asserted "args[1] is the verb; the surface id and payload follow as separate argv entries", describing a caller that does not exist.

A test that documents an imaginary caller is worse than no test, because it converts an untested path into one that looks covered. The allowlist assertions were all *correct* — every one of them — and the feature was still 100% broken for the only caller in the tree. That is a failure mode I will be watching for elsewhere in this codebase now.

The provenance note (`e32a6dd` landing after `3d6d52d`) is what turns it from "a bug" into "a class of bug": a guard added later, tested against the shape its author imagined rather than the shape already in the repository.

## On the fix itself

Taking the first whitespace token is the right call for the reason you gave, and it is worth restating because it is not the obvious one. This is not a loosening of the allowlist — `pipe-server.ts` `handleV1()` already parses the command as the first token, so before this change the allowlist and the parser it guards disagreed about what a request *was*. Now they agree. `"report_pwd notify"` being a `report_pwd` whose surface id is `notify` is not a concession, it is what the server was always going to do with it.

Scraping the verbs out of the shipped shell script so a new report cannot skip the check is the detail that keeps this from regressing.

And the silence is the reason it lasted: `>/dev/null 2>&1 &` meant the CLI exited 1 into a void, while `ping` and `list-surfaces` stayed healthy because the transport was never the problem. Anyone debugging it from the sidebar would have gone looking at the pipe.

## Housekeeping

`resources/cli/wmux.js` regenerated correctly and `npm run verify:resources` is clean on master. Good catch shipping it — a source-only fix here would have shipped nothing, since that file is what actually executes.

Your read on the six earlier failures was right too: `shell-context-menu`, `pty-ledger` and `pty-manager` are Windows-only suites that cannot run in a Linux devcontainer. Full suite on master after the merge: **1107 passed**.

PR #172, opened 2026-08-15
fix(pty): resolve Windows App Execution Aliases before spawn

### LINKED ISSUE #173: Windows: PowerShell 7 Preview profile fails with File not found: (empty path)
See #172. Importing WT's dynamic PowerShell 7 Preview profile and opening it throws `Failed to create terminal: File not found: ` (empty path). node-pty cannot spawn the `pwsh-preview` App Execution Alias.

bchap1n:
Fixes #173.

## Decision

Import Windows Terminal dynamic PowerShell Core profiles as `pwsh` / `pwsh-preview`, then resolve that name to a **real file** before `pty.spawn`. node-pty cannot spawn Windows App Execution Aliases.

## Why

WT's **PowerShell 7 Preview** profile has `source: Windows.Terminal.PowershellCore` and no `commandline`. Import used to leave it shell-less. Mapping it to `pwsh-preview` still crashed:

```
Failed to create terminal: File not found:
```

(empty path). `where pwsh-preview` finds the 0-byte WindowsApps alias. node-pty's SearchPath concatenates PATH + the bare name (no PATHEXT) and `GetFileAttributes` fails, so `get_shell_path` returns empty.

Preview on this machine is Store-only: `WindowsApps\Microsoft.PowerShellPreview_*\pwsh.exe`.

## Scope

- `mapWindowsTerminalProfiles`: dynamic PS Core → `pwsh` / `pwsh-preview` by name
- `resolveExistingShellPath`: first `where` hit that `existsSync`s; Preview falls back to the Store package `pwsh.exe`
- `resolveShell` / default shell use that path so already-saved `pwsh-preview` profiles work without re-import
- Extra args stay attached when the **requested** shell resolved (not when we fell back)

## Out of scope

Other dynamic WT profiles (WSL, Azure, VS Dev Prompt) stay shell-less and still fall back to the default shell.

## Verify

- `npx vitest run tests/unit/config-loader.test.ts tests/unit/pty-manager.test.ts` (48 passed)
- Open imported **PowerShell 7 Preview** on Windows with Store Preview installed — PTY starts, `$PSVersionTable` is 7.7.x preview

### DIFF
--- a/src/main/config-loader.ts
+++ b/src/main/config-loader.ts
@@ -24,6 +24,8 @@ interface WTProfile {
   name?: string;
   commandline?: string;
   startingDirectory?: string;
+  /** WT dynamic-profile source (e.g. "Windows.Terminal.PowershellCore"). */
+  source?: string;
   hidden?: boolean;
   font?: { face?: string; size?: number };
   fontSize?: number;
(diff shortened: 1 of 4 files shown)


amirlehmam:
Merged and shipped in [v1.2.0](https://github.com/amirlehmam/wmux/releases/tag/v1.2.0). Full write-up in #173 — including three adjacent defects this change surfaced, one of which (a Store-only *stable* PowerShell falling silently back to 5.1) is your bug one package over.

Thank you for sending the mechanism along with the fix. `get_shell_path` returning an empty string rather than naming what it could not find is the reason this was hard to read from the outside.

Issue #150, opened 2026-08-09
bug(pty): uncaught Napi::Error from conpty.node aborts the main process — 12 crashes, one signature (0.10 → 0.41)


Ray0483:
## Environment

- wmux: crashes observed on 0.10.0, 0.15.0, 0.17.0, 0.20.0, 0.25.0, 0.27.2, 0.29.1, 0.30.0, 0.38.0, 0.39.1, 0.41.0 (currently on 0.46.0)
- Windows 11 Pro 26200 (10.0.26200.2.0.0.256.48), x64
- Chromium/ANGLE in this build: `2.1.27949 git hash: bbf3d8a47552`
- Bundled `node-pty`: 1.1.0 (`resources/app.asar.unpacked/node_modules/node-pty`), ConPTY backend

## Symptom

wmux disappears instantly — the entire window, every workspace, every pane, with no error dialog and nothing written to the sidebar. It reads like a process kill, not a UI fault. Roughly once every 2–4 days under normal use.

Windows records it as `Application Error` / `BEX64`:

```
Faulting application name: wmux.exe, version: 0.41.0.0, time stamp: 0x6a3b18dc
Faulting module name: wmux.exe
Exception code: 0xc0000409          (STATUS_STACK_BUFFER_OVERRUN)
Fault offset:   0x0000000003290be5
Additional parameter: 0x7           (FAST_FAIL_FATAL_APP_EXIT)
```

`0xc0000409` here is not memory corruption — parameter `7` means the process deliberately called `__fastfail(FAST_FAIL_FATAL_APP_EXIT)`, i.e. the CRT `abort()` path.

## Evidence: 12 crashes, one signature

The Windows Event Log records 12 crashes between 2026-06-27 and 2026-08-04. I have minidumps from `%LOCALAPPDATA%\CrashDumps` for the 8 most recent, and every one is identical:

| Crash time (local) | wmux version | Exception | Param | Fault offset | Crashing thread |
|---|---|---|---|---|---|
| 2026-07-19 15:40:37 | 0.25.0 | 0xc0000409 | 0x7 | wmux.exe+0x3290be5 | index 0 (main) |
| 2026-07-22 14:04:18 | 0.27.2 | 0xc0000409 | 0x7 | wmux.exe+0x3290be5 | index 0 (main) |
| 2026-07-22 14:05:48 | 0.29.1 | 0xc0000409 | 0x7 | wmux.exe+0x3290be5 | index 0 (main) |
[...]

PR #178, opened 2026-08-17
Add multiple languages to app UI including pt, de, ja, hi, ru, uk, pl, tr, sv, nl, and cs

simplyBarbe:
## Summary

- Adds 11 new bundled UI languages to the Settings → General → Interface language dropdown: German, Portuguese, Japanese, Hindi, Russian, Ukrainian, Polish, Turkish, Swedish, Dutch, and Czech — bringing the app from 6 to 17 shipped languages (`en, fr, es, de, pt, it, nl, pl, tr, ru, uk, zh, ja, ko, hi, sv, cs`), catching the Electron app up to what `site/i18n.js` already ships on the marketing site.
- Each dictionary carries full 466/466 key coverage matching `en.ts` exactly (same key set, same order), registered in `BUILTIN` in the same relative order as `site/i18n.js` rather than alphabetically, with a comment explaining why.
- `de` had been the repo's go-to "unsupported/user-defined language" fixture across `user-locales-merge.test.ts` and the docs examples; since German is now bundled, those examples were moved to Tatar (`tt`), a language wmux will realistically never ship, so the fixture stays meaningful.
- Placeholder tokens (`{count}`, `{name}`, `{binding}`, etc.) were checked programmatically against `en.ts` for every language before commit — zero mismatches across all 11 additions.


amirlehmam:
Merged, and shipped in **1.4.0**. This is the most carefully prepared contribution this repo has had.

## What I verified before merging, since nothing else would have

There is no CI on pull requests here — the release workflow only runs on tags — so a PR of this size arrives entirely unchecked. Your summary made claims that were checkable, so I checked them rather than taking them:

- **466/466 key coverage.** Confirmed for all 17 bundled locales: no missing keys, no extra keys, no duplicates within a file. Every one is exactly `en.ts`'s key set.
- **Placeholder tokens.** Confirmed against `en.ts` for every key in every language — `{name}`, `{count}`, `{binding}`, all of them. Zero mismatches, as you said.
- **Values identical to English.** 23–51 per language, and I looked at what they are rather than counting them: `Terminal`, `Browser`, `Version`, `Markdown`, `Diff`, `Actions`. Cognates, not untranslated strings. The distribution is also the right shape — ~50 for the Latin-script languages, ~25 for `ja`/`ko`/`zh`.
- `npm test`, `tsc` on both projects, and `lint` are clean, with the lint problem count unchanged from master.

## The thing I most expected to be wrong, and wasn't

1.3.0 added `shortcutAction.resetTerminal` to `en.ts` after you opened this branch. The obvious failure mode for a 6,000-line translation PR is shipping a key set that was complete when it was written. You resynced, and all eleven new dictionaries carry it — translated, not stubbed:

```
de  'Terminalstatus zurücksetzen'
ru  'Сбросить состояние терминала'
ja  'ターミナル状態をリセット'
cs  'Obnovit stav terminálu'
```

That is the difference between a PR I can merge and one I have to send back.

Two other judgement calls I want to name because they were right and easy to get wrong. Moving the `de` fixture to Tatar rather than leaving a now-bundled language as the "unsupported language" example — the fixture only means something if wmux will realistically never ship it, and you noticed that shipping German broke the premise of tests you weren't otherwise touching. And ordering `BUILTIN` to match `site/i18n.js` instead of alphabetically, **with a comment saying why** — the next person to open that file will not "fix" it into alphabetical order now.

## Two corrections I made on top, in a separate commit

Kept separate from your commits so they stay attributable.

**German closing quotes.** All 11 quotations opened with `„` and closed with a straight ASCII `"` — `„Läuft"` rather than `„Läuft“`. None were correct, so it read as one systematic slip rather than a few typos. What made it stand out is that the harder cases are all right in this PR: Polish `„…”`, Czech `„…“`, Russian `«…»`, Japanese `「…」`. Locale-correct quotation marks are the detail most translation PRs skip entirely.

**Language count.** `CLAUDE.md` now said the site ships 22 languages; it ships 21 — `en fr es de pt it nl pl tr ru uk ar zh ja ko hi vi th id sv cs`. The RTL-Arabic half was right (`RTL_LANGS = ["ar"]`).

Neither is worth a round trip on a PR of this quality; both are worth being accurate about in the file that is supposed to be the source of truth for the next contributor.

## Not yours

While stress-running the suite I hit an intermittent `Worker exited unexpectedly` — about one run in three, always in the PTY tests, never in i18n. I chased it far enough to rule your changes out and to rule out the crash class in #150 (no `Application Error` in the Windows Event Log, so it is not the `0xc0000409` abort). It looks like node-pty's `conpty_console_list_agent.js` throwing `AttachConsole failed` inside a console-less vitest worker. Pre-existing, tracked separately, and mentioned here only so it does not look like it arrived with this PR.

Thank you. Seventeen languages, and the app has now caught up with the marketing site.

---

**1.4.0 is out** — https://github.com/amirlehmam/wmux/releases/tag/v1.4.0

Issue #125, opened 2026-07-28
updater broken in windows?


arnonuem:
It seems that windows does not have access to the sophisticated inline updater which pulls the binary itself and applies it.

When i click the button i am just redirected to the github repo where i have to download the installer myself and have to manually install it.

Is this intendet behaviour?

<img width="228" height="78" alt="Image" src="https://github.com/user-attachments/assets/d3e32f8e-de0b-4fa4-811e-ba90a76c2a24" />

amirlehmam:
No, that was not intended — and it's fixed in [v0.38.0](https://github.com/amirlehmam/wmux/releases/tag/v0.38.0).

Windows was never missing the updater. electron-updater has been running the whole time; it just had no UI. The badge you clicked came from a *separate*, notify-only GitHub poll whose only action was ever `openExternal(release page)`, so from the outside it looked exactly like "this platform doesn't get real updates". Two update paths, and the one you could see was the one that couldn't install anything.

The badge now drives the real updater:

- **Click** → checks, then downloads in place. The badge shows the percentage while it runs.
- **When it's ready** → the badge turns amber and reads `Restart`; clicking it (or accepting the dialog) restarts into the new version.
- **If this build genuinely can't self-update** — a dev run, `WMUX_DISABLE_UPDATER=1`, or a release published without `latest.yml` — it falls back to opening the release page, same as before. That path is now the exception rather than the only behaviour.

One deliberate detail: clicking the badge **skips the 3-day quarantine window**. That window exists to stop a malicious release from installing itself unattended before it can be spotted and yanked (#29) — it is not meant to stop *you* from installing a version you just read the number of and clicked. The unattended path still waits it out unchanged, and the install still needs your confirmation. Without this, a freshly published release would have shown a badge that did nothing for three days, which is the bug you filed wearing a different hat.

Note that the update *from* 0.37.0 still has to be done by hand — the fix ships in the version you're updating to. From 0.38.0 onward the badge does it for you.

PR #184, opened 2026-08-20
feat(updater): in-place updates for portable zip installs

2more-ger:
## Summary

In-place updates for **portable zip extracts** — the README-recommended install. NSIS installs are untouched.

Follow-up to #96 and #125. Those shipped an in-app path through `electron-updater`'s `NsisUpdater`, which can replace an NSIS install. A zip extract (no `Uninstall wmux.exe` next to `wmux.exe`) still cannot: listing the zip in `latest.yml` is the endless "update ready" loop from #96, and listing `setup.exe` would install a *second* copy under `%LOCALAPPDATA%` while leaving the zip folder on the old version. Clicking the badge then fell back to the GitHub release page (#125).

This PR gives zip extracts their own download → extract → swap path, driven by the same badge click and a new **Check for updates** button next to the version in Settings → Help.

## Zip only

Detection is conservative, so NSIS behaviour does not change:

| Layout | Marker | Updater |
| --- | --- | --- |
| NSIS (`wmux-*-setup.exe`) | `Uninstall wmux.exe` next to `wmux.exe` | existing `NsisUpdater` (quarantine, `latest.yml`, `quitAndInstall`) |
| Portable zip (`wmux-*-win-x64.zip`) | `wmux.exe`, **no** uninstaller | this PR: GitHub zip asset, extract, robocopy over the install folder after exit |
| `npm run dev` / `WMUX_DISABLE_UPDATER=1` | unpackaged or kill switch | still `handled: false` → GitHub releases page |

`initAutoUpdater()` returns before registering `NsisUpdater` on a zip extract, so a portable install can never enter the #96 loop. Zip updates are **user-initiated only** (badge / Help button). There is no unattended zip download; the 3-day quarantine window continues to apply only to the NSIS path.

## Runtime tools (no curl, no extra npm deps)

The download does **not** use `curl.exe` or `Invoke-WebRequest` (the latter's progress callback is what made large zips crawl). Packaged Electron already has Chromium networking.

| Step | Tool | Why it's always there |
| --- | --- | --- |
| Download | `electron.net.request` | Chromium, inside the packaged app |
| Extract | `%SystemRoot%\System32\tar.exe`, then Windows PowerShell `Expand-Archive` | `tar.exe` ships with Windows 10 1803+, which Electron 43 already requires; PowerShell is the fallback |
| Apply | `cmd.exe` + `robocopy` / `tasklist` / `timeout` / `findstr` via `%SystemRoot%\System32` | not looked up on `PATH` |
| MOTW | `Unblock-File` | best-effort; a missing/constrained PowerShell does not block relaunch |

Integrity: GitHub `asset.size` plus `digest` (`sha256:…`) when the API sends it. The payload is rejected if it does not contain `wmux.exe`.

## UI

- Titlebar badge: same click as today. Zip extracts no longer bounce to GitHub when `latest.yml` is an NSIS artifact.
- Settings → Help: button beside the version — *Check for updates* / *Update to vX* / download percent / *Restart*.

## Tests

- `tests/unit/zip-updater.test.ts` — portable vs NSIS detection, win-x64 asset pick, payload root, helper script.
- `tests/unit/updater.test.ts` — zip extracts do not call `NsisUpdater`; already-current zip reports `no_update`; NSIS path unchanged.

`npx vitest run tests/unit/updater.test.ts tests/unit/zip-updater.test.ts` — 36 passed. `npm run typecheck` clean.


amirlehmam:
Merged, and shipping in **v1.7.0** (release build running now).

Thanks — this is the piece #96 and #125 were missing. Those two got the *badge* right but left the README install, the one the docs actually tell people to download, as the one layout that couldn't self-update. Routing by install layout instead of trying to make `NsisUpdater` swallow a zip is the right call.

### What I verified before merging

- **The detection marker.** `isPortableZipInstall()` is the whole contract here — misfire it and an NSIS install gets a zip payload robocopied over it with a stale uninstaller and stale registry entries. I traced `Uninstall wmux.exe` back through the electron-builder we actually ship with rather than the docs: `common.nsh` defines `UNINSTALL_FILENAME "Uninstall ${PRODUCT_FILENAME}.exe"`, `PRODUCT_FILENAME` is `appInfo.productFilename`, which falls back to the sanitized `productName` since we set no `executableName`. `productName` is `"wmux"`, so the constant is exact. I've written that derivation into `CLAUDE.md` next to the file, because the name silently moves if either field is ever renamed.
- **The artifact side.** CI zips `release/win-unpacked/*`, so `wmux.exe` lands at the zip root — `findPayloadRoot()`'s direct case. Name is `wmux-${version}-win-x64.zip`, which `pickZipAsset()` matches on the preferred `win-x64` branch. Refusing to fall back to a marker-less `.zip` is a good instinct; that's exactly where a source archive would have slipped in.
- 1252 tests green, `tsc --noEmit` clean, and ESLint identical to the master baseline (29 problems both before and after — nothing new).

### One thing I changed on merge

`c35884c` — the helper's robocopy failure branch.

```
'if %ERRORLEVEL% GEQ 8 exit /b 1',
```

`applyStagedPortableUpdate()` calls `app.quit()` the moment the helper is detached, so by the time robocopy runs wmux is already gone. On `>= 8` the helper exited and **nothing came back** — the user clicked "Install and restart" and their multiplexer just disappeared, with no window and no error to read. That's reachable without anything exotic: a portable folder somewhere non-writable, or a leftover child still holding a DLL past `/R:5 /W:1`.

The failure branch now falls through to the same relaunch as the happy path, skipping only the MOTW strip:

```
'if %ERRORLEVEL% GEQ 8 goto relaunch',
'if exist "%SYS%\WindowsPowerShell\v1.0\powershell.exe" ... Unblock-File ...',
':relaunch',
'start "" "%EXE%"',
```

A failed copy usually leaves the old build in place, so `%EXE%` still starts — on the old version, which beats not starting. Test added in `zip-updater.test.ts`.

Same commit flattens `downloadToFile()`'s stream callbacks and `runHidden()`'s error string; our sonarjs gate rejects nesting past 4 levels and nested template literals, and the new file tripped both. Pure refactor, behaviour identical — the size/sha256 checks moved into a `verifyDownload()` helper.

### Follow-ups, not blockers

- **i18n.** The 4 new `settings.help.*` keys landed in `en` and `de` only. `translate()` chains through English, so the other 16 locales render readable English rather than raw keys — degrades fine, but they're still untranslated. I deliberately didn't machine-translate them into 16 languages on your behalf; happy to take a follow-up PR, or they'll get picked up the next time someone sweeps the locales.
- **robocopy has no `/PURGE`.** Files dropped in a newer version linger in the install root instead of being removed. Almost certainly the right default — `/PURGE` over a folder a user may have put things in is how you delete someone's data — but worth a comment so the next reader knows it's a choice and not an oversight.
- **Backpressure in `downloadToFile()`.** `out.write(chunk)`'s return value is ignored, so a slow disk buffers the whole ~100MB in memory. Disk normally outruns network here, so it's theoretical; a `res.pause()`/`out.once('drain')` pair would close it if it ever shows up.

One note for anyone reading this later: nobody on <= 1.6.0 gets the new path, since the code doing the swap ships *in* the update. Zip extracts have to take one more manual trip to the releases page to reach 1.7.0, and self-update from there on.

PR #185, opened 2026-08-21
i18n(zh-TW): translate the 18 new Saved Layouts keys

Ray0483:
Follow-up to #180.

The saved-layouts feature added 18 keys to `en.ts` that no locale has translated yet, so
zh-TW users see English in **Settings → Workspace → Saved Layouts** and in the
**Layouts** category of the command palette. This PR fills them in.

**Scope:** `src/renderer/i18n/locales/zh-TW.ts` only, **+18 lines, 0 changed lines.**
No existing translation is touched, and every added key already exists in `en.ts`, so
the `Translation` type and the stale-key check in `tests/unit/i18n.test.ts` are unaffected.

Terminology follows what is already in the file (and Taiwanese software convention):

| en | zh-TW |
|---|---|
| pane | 窗格 |
| layout | 版面 |
| preset | 預設集 |
| default | 預設 |
| command | 指令 |

FWIW the other 16 locales are missing the same 18 keys — they currently fall back to
English, which is harmless, but it may be worth a sweep.


amirlehmam:
Merged — thanks.

Verified parity rather than eyeballing the diff: zh-TW is back to **488/488** against `en.ts`, no duplicate keys, and no key that doesn't exist in the English dictionary (a typo'd key is the failure mode that matters here — `translate()` falls back through English, so a misspelled key renders English forever and nobody notices it's untranslated rather than deliberately kept).

Terminology reads consistent with the rest of the file — 窗格 for pane, 工作區 for workspace, 版面 for layout — and `{n}` / `{name}` placeholders are intact in all three interpolated strings, which is the other thing that silently breaks.

Shipping in the next release.

Issue #89, opened 2026-07-11
Backgrounds parallel to theming


arnonuem:
Since you asked for it, what i liked in wave was the possibility to set a custom background that has nothing to do with the theming. And what also would be an UX improvement -> setting the font with a font dialog that finds available fonts in the system. With freetext fields i am uncertain what to type to swap the font with my favourite one. Ok thats maybe two issues in one :D

like this:

```
{
  "bg@mylovelybackground": {
    "display:name": "MyLovelyBackground",
    "bg": "radial-gradient(ellipse at 0% 0%, rgba(9, 140, 206, 0.40) 0%, transparent 75%), radial-gradient(ellipse at 100% 100%, rgba(137, 33, 210, 0.35) 0%, transparent 75%), #1a1a1a",
    "bg:opacity": 0.95,
    "bg:*": true
  }
}
```

amirlehmam:
Both halves of this shipped in [v0.24.0](https://github.com/amirlehmam/wmux/releases/tag/v0.24.0) 🎨

**Custom backgrounds parallel to theming** — Settings → General → *Custom background*:
- Paste any CSS `background` (gradients, colors, `url(…)` images) into the free-form field — your example works as-is. There's also a preset dropdown, and the first preset ("Lovely") is literally the gradient from your post.
- A **Terminal opacity** slider (your `bg:opacity`) controls how much of the background shows through the terminal color scheme — the terminals render with a translucent theme background on top of the custom layer, so it composes with any color scheme.
- Live preview right in the settings panel.

**Font picker** — Settings → Terminal now has a dropdown listing every font family installed on your system (enumerated from the Windows font registry), with a live preview line so you can see the pick immediately. The free-text stack field is still there below it for advanced fallback stacks.

Grab the zip from the release (right-click → Unblock after download), or wait for the in-app update badge. Thanks for the great suggestion — the Wave-style config example made the intent crystal clear!

arnonuem:
just wow 🥳

arnonuem:
Not yet working as intended. I guess it has to do with the theming interfering with the background. The preview looks ok

<img width="564" height="204" alt="Image" src="https://github.com/user-attachments/assets/1ccd7d79-650b-407c-8456-172dfc26878b" />

And i can see the background behind the terminals -->

<img width="857" height="829" alt="Image" src="https://github.com/user-attachments/assets/bb545fca-7630-4abe-9d56-6903c2909647" />

[...]

PR #192, opened 2026-08-24
Add support for terminal transparency and related terminal profile imports

jeFF0Falltrades:
Lots of code changes, but a pretty straightforward feature: I just wanted to replicate the transparency/opacity settings seen in other terminals (including cmux) that allows for some customization of how much of the background comes through the terminal.

Since wmux is Windows-focused, I also added some support for `acrylic` and `mica` materials from Win11 for more customization of the look and feel.

Lastly, I also added functionality to import terminal settings from Ghostty and Windows Terminal (as well as an `undo` switch on import so no settings are lost until they are actually saved).

Bot description of changes follows - Thanks!

---

## Summary 
Adds real window transparency to `wmux`: seeing the actual desktop through the terminal, not just a background layer inside an opaque window.

wmux already had a `terminalBgOpacity` slider, but it only ever applied over the in-app custom background (issue #89). 
The window itself stayed opaque, so the thing most people mean by "terminal opacity" wasn't possible at all.

### Three styles

| Style | What it does | Requires |
|---|---|---|
| **Clear** (default) | Plain per-pixel alpha, no blur — read a browser through the window, the way Windows Terminal does it | Windows 10+ (DWM) |
| **Acrylic** | Strong frosted blur | Windows 11 (22000+) |
| **Mica** | Subtle desktop tint | Windows 11 (22000+) |

`Clear` is the default because it's the one that matches the expectation. 

Acrylic and Mica are DWM *backdrops*, and a backdrop blurs what's behind it by definition — no setting of them could ever let you read text through the window.

On Windows 10 the blur materials are offered but disabled rather than omitted. 

Chrome stays opaque on purpose — the sidebar and titlebar keep `--ui-bg-0` so their text stays readable over an arbitrary wallpaper.

### Config import, now actually reachable

`ThemeConfig.backgroundOpacity` was parsed in two places and read in none. The cause turned out to be bigger than a missing lookup: **the entire theme-import path was dead code.** `parseWindowsTerminalConfig` and `parseGhosttyConfig` had IPC handlers and preload bindings, and nothing in the UI had ever called either — so the field had no way to reach the renderer even in principle.

**Settings → Terminal → Import** is the missing end of that path. It applies the parsed colors as a custom scheme, selects it, takes the font, and seeds `terminalBgOpacity`. 

Three parser bugs fixed on the way, all of them only reachable once something finally called this code:

- **WT opacity wasn't parsed at all** — `schemeToTheme` hardcoded `1.0`. Now read from both spellings: modern `opacity` (0-100, applies whether or not acrylic is on, since `useAcrylic` only controls blur) and pre-1.12 `acrylicOpacity` (0-1, ignored by WT unless `useAcrylic` is set). A profile carrying both predates the rename and has been half-migrated, so the modern key wins.
- **`profiles.defaults` was ignored.** WT puts inherited settings there, and most real configs keep `colorScheme`, `font` and `opacity` in it rather than on the profile. Tested against a real settings.json: before, it imported `Cascadia Mono / 13 / opacity 1.0`; after, `SynthWave 84 / #262335 / Fira Code / 10 / 0.8`. Colors had only survived by luck — the `schemes[0]` fallback happened to land on the right one.
- **Ghostty's `background-opacity = 0` imported as fully opaque.** `parseFloat('0') || 1.0` — of every value the key can hold, the one silently discarded was the one you could only have written deliberately.

### One bug, five times: everything that only looked right behind an opaque window

Turning the window transparent exposed a whole class of elements that had never actually painted. Each was invisible while something opaque sat behind it.

**The background was painted twice.** `.terminal-pane__container` and xterm's own inline background on `.xterm-scrollable-element` both painted the same rgba, one over the other. Two identical translucent layers resolve to `1-(1-a)²`, not to their own alpha — so 80% rendered at 96% and 50% at 75%. This was the whole reason wmux looked stubbornly more opaque than WT at the same number.

**Clear mode came up solid white.** `transparent` is ignored on Windows unless the window is frameless, and `titleBarStyle: 'hidden'` is *not* frameless. An ignored `transparent` doesn't fall back to the theme either; Electron's default `backgroundColor` is `#FFF`. Clear windows are built with `frame: false` now, and the renderer draws its own caption buttons in the 140px the overlay used to occupy.

**Four surfaces in the terminal area never painted.** With panes at 80% and the gaps at 0%, a split reads as bright seams: `.xterm`'s 2px padding (filled per-pane, from the colour that pane actually took — a pane with a `--color-scheme` override has to match itself), the 6px `.split-divider` and `.pane-wrapper`'s 1px border (filled from the global theme, since a divider lies between two panes that may disagree), and the **focused pane's accent ring**.

**Gutters could collapse to nothing.** With a custom background `terminalBgAlpha` returns 0 by design — that layer becomes the surface — so `--wmux-pane-fill` became `rgba(…, 0)`. Right for a border painted *on* that layer, wrong for a 6px divider, which became a hole through the window. `--wmux-pane-gutter` is the same colour with one difference: it is never nothing.

**And the browser panel's resize handle had no background at all.** The browser panel isn't in the split tree — it's a separate right-hand panel with its own hand-rolled 4px handle carrying a cursor, a mousedown handler and nothing else. Invisible over an opaque window; a 4px hole over a transparent one.

It gets the dividers' shape but not their number, which is the part worth flagging as a trap rather than a duplicate. A divider runs between two surface tab bars and is opaque for their 28px. This handle runs between a tab bar and the browser's own header, which is 34px — so reusing the divider's constant left a 4×6px notch of desktop beside the address bar. It takes the taller of its two neighbours: the sides disagree, and only one of the two wrong answers is visible. Too much chrome reads as the header extending 4px; too little reads as a hole punched through the window.

**`needsRestart` was answered fleet-wide.** `setBackdrop` applies to every window and was aggregating one boolean across all of them, which the IPC handler returned to whichever window asked. Opening a second window while a change was pending gave it a restart banner for something already done — and made its renderer keep treating itself as opaque, refusing to paint the transparency it had just been correctly built with. Scoped to `e.sender` now.

### Design decisions worth flagging

**Opacity floors at 15%.** 0% was reachable and it's too far — in Clear mode nothing blurs behind the terminal, so a fully transparent background leaves text floating on the desktop with nothing to sit on. 15 is borrowed from Ghostty, which enforces exactly that on `unfocused-split-opacity`: *"because 0 is not useful (it makes the window look very weird), the minimum value is 0.15."* Worth being straight that the precedent is narrower than it looks — neither Ghostty's `background-opacity` nor WT's `opacity` clamps at 0 in config. Only the reasoning is borrowed; the floor is ours. It applies on read, not on write, so a value imported from a WT profile survives intact if the floor ever moves.

**A custom background replaces the terminal colour** rather than sitting behind it. Previously the terminal's own colour was drawn opaque over the custom layer, so at 100% opacity the background was perfectly hidden — setting a background and seeing no background. Both references treat it the other way round (Ghostty composites `background-image` over the colour, WT draws `backgroundImage` at full strength). This is a visible behaviour change for anyone who had a custom background *and* a tuned opacity: they now get the background unblended. The old look is still reachable — `customBackground` takes raw CSS, so a translucent gradient composites the same way.

**Import is undoable.** One click writes seven settings across two slices and raises a restart banner; someone aiming at the button beside it loses their font size with no record of what it was. Each import parks the exact keys it's about to overwrite and offers them back. The snapshot takes `userColorSchemes` as a whole map, because the source may name a scheme the user already had and the import replaces it in place.

Only one import runs at a time — both buttons ran unawaited, and two in flight would each read prefs the other hadn't written yet. That guard lives at module scope rather than in a ref, for the same reason the undo snapshot lives in the store: this panel unmounts on a tab switch, so anything held in component state dies exactly when the user clicks over to General to look at the restart banner the import just raised.

Deliberately session-scoped, not persisted: an undo that outlived the session would offer to revert a font size the user had since changed on purpose. Both reference implementations agree — WT scopes Discard Changes the same way (only until you save), and cmux reads Ghostty's config in place rather than copying it, so it has nothing to undo at all.

### Files

New:
- `src/renderer/store/backdrop.ts` — one shared definition of the alpha calculation, which had accumulated three drifting copies across `useTerminal`, `usePaneFill` and `App`. Any drift renders as a seam at the edge of a pane.
- `src/renderer/hooks/useWindowTransparency.ts` — drives the two halves (main-process window flags, renderer `wmux-transparent` class) that have to agree
- `src/renderer/hooks/usePaneFill.ts` — publishes `--wmux-pane-fill`, `--wmux-pane-gutter` and `--wmux-pane-ring` for the chrome between and around panes
- `src/renderer/components/Titlebar/WindowControls.tsx` — renderer-drawn caption buttons for frameless windows
- `src/renderer/utils/backdrop-caps.ts` — the host capability, fetched once

### Testing

34 new unit tests (15 in `backdrop-alpha.test.ts`, 19 added to `config-loader.test.ts`) pinning the alpha calculation, the floor, the custom-background composite, WT's two opacity spellings, `profiles.defaults` inheritance, and Ghostty's opacity parsing including the zero case.

Suite goes from 1250 to 1284 passing. The 38 pre-existing failures are unchanged and identical on master — Windows-path and PowerShell-dependent tests that can't run under WSL. No new lint errors; the 16 in `src/` are present on master file-for-file.

Verified live on Windows 11 against a real Windows Terminal SynthWave 84 profile, and the remaining artifacts were closed by sampling the rendered pixels rather than by eye — the fixed gutter measures `(37,37,37)` with zero variance down the full pane height where the desktop previously showed through.

### Known limitations

**An already-open second window doesn't follow a transparency change until it's restarted.** `setBackdrop` applies to every window, but `setAppearancePrefs` only updates the calling window's own Zustand store and persists to disk — wmux has no cross-window settings broadcast, for this or any other setting. A window opened *after* the change reads `settings.json` at creation and comes up correct; one already open keeps its old store.

Two consequences worth naming. A material-only change (acrylic→mica, which needs no restart) *is* applied to the sibling's real window by that same loop, so the OS repaints it while its Settings panel still shows the old material. And because `setAppearancePrefs` persists the whole merged prefs object, a later unrelated appearance change made in the stale window writes its stale transparency values back over the new ones on disk.

This is a pre-existing gap rather than something this PR introduces — change the font size in one window today and the other keeps the old one, with the same last-writer-wins behaviour on disk. Closing it means main pushing pref changes into renderer stores, which is a repo-wide decision about where settings truth lives and wants its own PR. Flagged here because this PR's fleet-wide `setBackdrop` is the first thing that makes the asymmetry visible.

Example `Clear` Display:

image

Example `Acrylic` Display:

image

Example `Mica` Display:

image


amirlehmam:
Merged as `4c139be`, shipping in **v1.11.0**. This is one of the best PRs this repo has had — thank you.

What sold it wasn't the feature, it was the second half of the description: turning the window transparent didn't *add* five bugs, it revealed five that had always been there and had simply never had to paint. The double-composited background in particular explains a complaint I'd never been able to reproduce — people saying wmux looked more opaque than Windows Terminal at the same number. `1-(1-a)²` is exactly right and I'd never spotted it: 80% rendering at 96%.

A few other things I appreciated:

- **`transparent` is ignored unless the window is frameless, and `titleBarStyle: 'hidden'` is not frameless.** That is a genuinely nasty Electron edge, and the failure mode (silent fallback to `#FFF`, not to the theme) is the kind that costs an afternoon.
- **`ThemeConfig.backgroundOpacity` parsed in two places and read in none.** You were right that the whole import path was dead code — IPC handlers and preload bindings wired to nothing. It had been that way for a long time and I'd have called it "a missing lookup" too.
- **`parseFloat('0') || 1.0`.** Of every value that key can hold, discarding the only one you'd have to write on purpose. Nice catch.
- `backdrop.ts` collapsing three drifting copies of the alpha calculation into one, and `toWindowMaterial()` narrowing untrusted input to the safe value rather than trusting the persisted string.

### Verified before merge, on Windows 11

- `npm run build:main` and `npx vite build` clean
- **1371 tests pass**, up from 1337 (your 34, as advertised) — for what it's worth, the 38 failures you saw are a WSL artifact; the Windows-path and PowerShell tests all pass on the real thing
- lint at 16 errors, identical to master file-for-file

### Two notes, neither blocking

**The multi-window limitation you flagged is the right call.** Making `setAppearancePrefs` push into sibling renderer stores is a decision about where settings truth lives, and it should not ride in on a transparency PR. Naming the last-writer-wins consequence on disk was the useful part — that's the bit that would have surprised someone later. It's pre-existing (font size behaves identically today), and I'd rather fix it once for every pref.

**i18n.** The new Settings strings landed in `en.ts` only. That's fine — `translate()` falls back `DICTIONARIES[lang] ?? DICTIONARIES.en ?? key`, so the other 17 locales render English rather than raw keys. Nothing to do on your side; I'll fold them into the next translation pass.

The pixel-sampling verification (`(37,37,37)` with zero variance down the pane) rather than checking by eye is the detail I'll remember. Genuinely appreciated.

PR #196, opened 2026-08-24
fix(pty): prefer Windows OpenSSH for a bare ssh spec

### LINKED ISSUE #193: Prefer Windows OpenSSH when resolving a bare ssh shell spec
## Summary

On Windows, a bare `ssh user@host` shell spec can resolve through `PATH` before the native Windows OpenSSH installation is considered. If Git or another package contributes an `ssh.exe` earlier on `PATH`, wmux can spawn that client even when the matching Windows `ssh.exe` and `scp.exe` pair is installed.

## Why this matters

- The selected client may not match the Windows OpenSSH behavior wmux expects.
- Features that need the matching `scp.exe` cannot safely infer it from an unrelated PATH client.
- A normal Windows installation should use `%ProgramFiles%\OpenSSH` when both tools exist, then `%SystemRoot%\System32\OpenSSH`, before falling back to `PATH`.

## Expected behavior

Resolution order for a bare SSH spec should be:

1. An explicitly requested executable path.
2. A complete OpenSSH pair under Program Files.
3. Windows System32 OpenSSH.
4. The normal PATH fallback.

The behavior should be covered by Windows-focused unit tests.

LevSky22:
## Summary

- prefer an explicit SSH executable when one was requested
- otherwise prefer a complete OpenSSH `ssh.exe`/`scp.exe` pair under Program Files
- fall back to Windows System32 OpenSSH before consulting `PATH`
- keep the resolver memoized and cover the Windows precedence cases

This avoids accidentally selecting Git's or another package's PATH-provided SSH client on Windows and gives later consumers a deterministic matching tool pair.

## Validation

- 44 focused unit tests passed
- `npm run typecheck`

Closes #193

### DIFF
--- a/src/main/pty-manager.ts
+++ b/src/main/pty-manager.ts
@@ -11,6 +11,7 @@ import { attachErrorSink, installPtyCrashGuard } from './pty-crash-guard';
 import { powerShellShimDir } from './powershell-shim';
 import { getCliBinPath } from './cli-paths';
 import { getNodeRuntime } from './node-runtime';
+import { opensshPath } from './system32';
 
 // Applied once, at load, before any PTY can exist — the exit callback it guards
 // is registered by node-pty inside pty.spawn(), so a later install would leave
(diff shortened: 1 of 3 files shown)


amirlehmam:
Merged. Thank you — this one is a genuinely good catch.

The part I want to call out is that you found the *right* place for it. `resolveExistingShellPathUncached` only ever sees a shell **spec** — something wmux is spawning itself — so an absolute path the user wrote still wins, and a command typed at a prompt is resolved by the shell inside the PTY and never reaches this function. That means panes keep behaving like every other terminal, and the fix is invisible to anyone whose keys live on disk. A change that intercepted `ssh` more broadly would have been the wrong trade.

Pulling `opensshPath()` out into `system32.ts` with an injectable `exists` was also the right call: it made the Program-Files-over-System32 precedence testable without a Windows fixture, and #198 then needed exactly that helper to pick the `scp.exe` from the same authentication stack.

Shipping in **v1.12.0**.

PR #197, opened 2026-08-24
fix(session): keep the requested shell spec through a remount

### LINKED ISSUE #194: Preserve the requested shell specification across terminal remounts
## Summary

When a terminal reports the executable that was actually spawned, wmux writes that resolved executable back into `SurfaceRef.shell`. That field is also the requested respawn/session-restore specification.

For a surface created with a command such as `ssh user@host`, the resolved value is only `...\ssh.exe`; the destination and other arguments live separately. After a React remount or persisted-session restore, wmux can therefore respawn bare `ssh.exe`, which prints usage and exits.

## Expected behavior

- Preserve `SurfaceRef.shell` as the requested shell specification.
- Store the executable used for labels/display in a separate optional field.
- Ensure remounts, persistence, labels, and old serialized sessions remain compatible.

This is independent of remote file upload; it affects any shell specification whose arguments must survive a remount.

LevSky22:
## Summary

- preserve `SurfaceRef.shell` as the requested respawn/session-restore specification
- store the actual spawned executable separately as `resolvedShell`
- use the resolved value only for the terminal label
- keep persisted sessions without the new optional field compatible

This prevents a shell such as `ssh user@host` from becoming a destination-less `ssh.exe` after a remount or restore.

## Validation

- 14 focused unit tests passed
- `npm run typecheck`

Closes #194

### DIFF
--- a/src/renderer/components/SplitPane/surface-label.ts
+++ b/src/renderer/components/SplitPane/surface-label.ts
@@ -37,7 +37,11 @@ export function getSurfaceLabel(
     case 'terminal': {
       const folder = surface.currentCwd ? cwdFolderName(surface.currentCwd) : null;
       if (folder) return folder;
-      return getShellLabel(surface.shell || workspaceShell) || t('surfaceLabel.terminal', 'Terminal');
+      // resolvedShell first: it is the concrete executable, so it labels a
+      // pane started with no spec at all. `shell` may be a whole command line
+      // (`ssh user@host`), which getShellLabel would render as a mouthful.
+      return getShellLabel(surface.resolvedShell || surface.shell || workspaceShell)
+        || t('surfaceLabel.terminal', 'Terminal');
     }
     case 'browser':
       return t('surfaceLabel.browser', 'Browser');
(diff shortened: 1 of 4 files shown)


amirlehmam:
Merged. Thank you.

This is the kind of bug I like seeing fixed properly: the symptom was "restored ssh workspace prints usage and exits", and the tempting fix is a guard at the respawn site. Splitting the field instead — `shell` stays the requested spec, `resolvedShell` carries the executable for the caption — fixes it at the cause, and it fixes it for *every* spec that carries arguments, not just ssh.

Narrowing `updateSurface` to `Partial<Omit<SurfaceRef, 'id' | 'type' | 'shell'>>` in #198 is the part that makes it stay fixed. The regression was reintroducible by anyone reaching for the obvious field name; now it is a type error.

I checked the persistence side too — `SessionData.workspaces[].splitTree` is typed `any`, so `resolvedShell` rides along without a schema change, and older `session.json` files without the field restore fine. That matches the invariant from #145 (persistence types must be a superset of what every writer emits).

Shipping in **v1.12.0**.

PR #198, opened 2026-08-24
feat(terminal): upload local files in direct SSH panes

### LINKED ISSUE #195: Upload pasted and dropped local files in direct SSH panes
## Summary

Pasting a screenshot, pasting an Explorer-copied file, or dropping a local file into a terminal connected over SSH currently inserts a Windows path that does not exist on the remote host.

Add an opt-out remote insertion path for direct SSH panes: upload local files with the matching Windows OpenSSH `scp.exe`, then insert safely quoted remote paths into the terminal. This is inspired by cmux's remote file handling, adapted for Windows process and terminal semantics.

## Proposed behavior

- Handle clipboard images, copied files, and file drops through one main-process resolver.
- Detect `wmux ssh` and direct SSH commands reported by integrated PowerShell/Bash shells.
- Preserve the pane's SSH destination, port, identities, supported options, working directory, and exact client installation.
- Upload into a private, unique `/tmp/wmux-drop-<batch-id>/` directory using `BatchMode=yes`.
- Keep plain text paste unchanged.
- Allow Shift-drop to insert the local path without uploading.
- Provide independent `upload-on-paste` and `upload-on-drop` settings, enabled by default.
- Fail closed when the active destination or matching SSH/SCP client cannot be verified.

## Windows safety boundary

Unlike macOS, Windows does not expose a reliable tty foreground process group. A descendant `ssh.exe` may be a background process, so process ancestry must only corroborate an authoritative managed or shell-integration report; it must never select an upload host by itself.

Nested SSH started after reaching the first remote host cannot be safely reproduced by the Windows client and is out of scope. Observable destination mismatches should block upload, and the limitation should be documented.

## Related foundations

- #193 fixes deterministic Windows OpenSSH client selection.
- #194 preserves requested shell arguments across remount/session restore.

LevSky22:
## Summary

- upload pasted screenshots, Explorer-copied files, and dropped files when a pane is directly connected over SSH
- preserve SSH destination, port, identities, supported connection options, working directory, and exact Windows OpenSSH installation
- use one main-process insertion resolver with per-surface ordering and cancellation
- place uploads in a private unique remote batch directory and insert safely quoted remote paths
- add independent paste/drop settings, Shift-drop bypass, translated failures, documentation, and generated resources
- harden renderer ownership, DOM `File` conversion, lifecycle report ordering, nested-session ambiguity, and configuration validation

The design follows cmux's user-facing behavior where Windows provides equivalent evidence. Windows has no reliable tty foreground process group, so process ancestry is deliberately corroborating evidence only and can never independently select an upload destination. Unknown or contradictory destinations fail closed.

## Dependencies

- #193 / #196 (`fix/ssh-path-resolution`)
- #194 / #197 (`fix/preserve-shell-spec`)

This PR is opened as a draft because its branch currently contains those two foundation fixes. Once they merge, the branch can be rebased and the resulting feature-only diff marked ready.

## Validation

- `npm test`: 1,530 passed, 11 skipped
- automated SSH/SCP upload tests cover argument fidelity, matching client selection, private remote-directory setup, multi-file ordering, cleanup, failures, and cancellation
- detector tests cover managed and shell-reported sessions, stale/out-of-order reports, nested ambiguity, and the Windows background-SSH fail-safe
- renderer/preload tests cover genuine DOM `File` conversion, surface ownership, queued insertion, remounts, and teardown
- `npm run typecheck`
- `npm run build:main`
- `npm run build:renderer`
- `npm run verify:resources`
- PowerShell integration parser check
- `git diff --check`

### Live end-to-end verification

The feature was also exercised through a running wmux/Claude Code terminal against a real SSH host using Windows OpenSSH:

- a generated 120×50 PNG placed on the Windows file-drop clipboard and pasted with `Ctrl+Shift+V` increased the remote upload count from 0 to 1
- the uploaded remote file was verified as the expected 120×50 PNG
- plain-text paste through the same binding left the upload count unchanged at 1 and still reached the terminal
- the remote temporary test file was cleaned up afterward

The subsequent v1.11 rebase and final private-directory, ownership, cancellation, and fail-safe hardening were covered by the automated tests above. That final hardened revision was not separately re-smoked against the host.

Repository-wide lint remains at the existing upstream baseline of 14 errors and 13 warnings; this branch adds no new lint findings.

Closes #195


amirlehmam:
Merged. I rebased the branch onto master myself once #196 and #197 landed (the squash-merge changed their SHAs, so the branch went `CONFLICTING` the moment I marked it ready) — the result was byte-identical to your head, and the PR then showed the feature-only diff you described: **+3774 −162 across 36 files**. Full suite green on the integration branch: 1,536 passed, 6 skipped.

This is an unusually well-argued contribution and I want to be specific about why I took it at this size.

**The Windows safety boundary is the whole thing, and you got it right.** The precedence rule in `ssh-detect.ts` — probe evidence may *corroborate* a managed or reported session but may never *establish* one — is not a nicety, it is the entire difference between this feature and a way to silently ship a user's screenshot to a background `ssh host &`. You documented why cmux can be laxer (macOS `ps -t <tty>` answers the foreground-process-group question; ConPTY has no tpgid), and then you actually held the line in `refresh()` instead of quietly relaxing it for the convenient case. Same for the nested-ssh block in `detect()` and the fail-closed `remoteHint` path.

Other things I checked and liked:

- `posixShellQuote` is unconditional single-quoting, with the reasoning stated. Conditional quoting on a filesystem whose conventions we don't know is how this goes wrong.
- `execFile` with an argv array throughout; `destination` never reaches a shell.
- The all-or-nothing batch with `rm -rf` rollback. A partial batch handing the user some remote paths and some silence is genuinely worse than a clean failure.
- `report_command` is filtered to `ssh` *in the shell integration*, not in main — so the pipe never carries the user's other command lines at all. Not forwarding it to renderers on top of that is the right belt-and-braces.
- Taking DOM `File` objects in `resolveDrop` rather than path strings. Accepting paths would have been an arbitrary local-file upload API for a compromised renderer.

**One follow-up I pushed on top** (`24f4554`):

`SshDetector.refresh()` awaited a full `Get-CimInstance Win32_Process` sweep before reading its layers — but by your own precedence rule, with neither `managed` nor `reported` present every branch below that await already returns `null`. So pasting a screenshot into an ordinary **local** pane paid ~550ms enumerating every process on the machine, and `start()` then held the sweep on a 3s timer for five more passes, to answer a question a local pane cannot ask. I short-circuited it in the detector (so every caller benefits) and stopped arming the probe from `resolveFor` unless `remoteHint` already names a destination. Behaviour is unchanged; it's purely the local path getting its latency back.

That change made your `never treats a probe-only descendant as the foreground ssh session` test pass vacuously, so I reworked it to give a *second* surface an authoritative report — the sweep genuinely runs and the background ssh is still refused, which is what the test was written to prove — and added a separate test asserting the sweep isn't called at all for a pane with no report.

Also in that commit: `terminal.uploadFailed` into the other 17 bundled locales (it shipped English-only, so every non-English user got the fallback string), plus four SonarJS findings cleared on the way past — the sequenced-report regex is now a linear hand-written split rather than `^(seq=\d+)\s+(.*)$` run over whole command lines, the `void` promise discards became explicit catches, and the ancestry walk moved out of `attributeSshProcesses` into `owningSurface`.

Two notes for anyone reading later, neither blocking:

1. The `BatchMode=yes` trade means key/agent auth only. That is the right default — there is no TTY for a passphrase prompt to go to — but it's worth remembering when someone reports "upload does nothing" on a password-auth host. The failure toast surfaces scp's own last stderr line, which should make that self-diagnosing.
2. `refresh()` returning `null` when the exact client can't be corroborated is strict enough that a CIM failure blocks upload on a pane that *is* remote. Fail-closed is correct here; I just want the reasoning on the record if it ever gets reported as a bug.

Shipping in **v1.12.0**. Thanks — this raised the bar.

amirlehmam:
Released: **[v1.12.0](https://github.com/amirlehmam/wmux/releases/tag/v1.12.0)**. CI green end to end — 1,536 tests, resource verification, electron-builder package, portable zip and `latest.yml` (pointing at the NSIS `setup.exe`, so installed clients update cleanly).

Issue #143, opened 2026-08-04
0.42.0: V2 commands resolve to the wrong window once a second window exists, and unknown flags silently execute with defaults


Ray0483:
**wmux version:** 0.42.0
**OS:** Windows 11 Pro 10.0.26200
**Shell in pane:** powershell.exe

Two CLI correctness problems found while verifying the #141 fix (which is fixed — details in a comment there). Both matter mainly to scripts and agents driving wmux from inside a pane.

---

## 1. `WMUX_SURFACE_ID` resolves to the wrong window once a second window exists

The 0.42.0 release note says *"V2 commands now use caller's `WMUX_SURFACE_ID` to identify owning window"*. With a single window this appears to work. **As soon as a second window exists, passing `WMUX_SURFACE_ID` makes commands resolve to a constant, wrong window — and the caller's own surface is not in the result.**

### Setup

- Caller shell lives in `surf-360768cf-…`, in pane `pane-c36f0826-…`, in workspace "Workspace 3", in window `win-fda7feb5-…`.
- A second window `win-2806cd6c-…` was opened with `wmux new-window`.
- There is also a workspace "Session 1" containing three terminal panes (`pane-f45f55c3`, `pane-b11e7cf2`, `pane-89504319`).

### Observed

| Focused window | Call | Panes returned by `tree` | Contains caller's pane `c36f0826`? |
|---|---|---|---|
| original | `wmux tree` (with `WMUX_SURFACE_ID`) | f45f55c3, b11e7cf2, 89504319 | ❌ |
| original | `wmux tree` (env var removed) | c36f0826, 48d68481 | ✅ |
| new | `wmux tree` (with `WMUX_SURFACE_ID`) | f45f55c3, b11e7cf2, 89504319 | ❌ |
| new | `wmux tree` (env var removed) | f45f55c3, b11e7cf2, 89504319 | ❌ |

Grepping the `WMUX_SURFACE_ID` result for the caller's own surface id returns **0 matches**. The result is byte-stable across focus changes: it always resolves to "Session 1", a workspace the caller is not in.

So the env var makes resolution *worse*: removing it at least follows the focused window.

`tree` and `list-surfaces` now agree with each other (that part of #141 is fixed) — they are simply wrong together.

### Reproduce

```bash
# from a shell inside a wmux pane, with at least one other workspace present
echo "$WMUX_SURFACE_ID"          # note this id
wmux tree                        # ✓ shows your pane
wmux new-window
[...]

PR #200, opened 2026-08-24
feat(cli): add caller-scoped `wmux current-workspace`

derekneely:
## What this adds

`wmux current-workspace` (alias `whoami`) — the workspace the calling pane is in:

```console
$ wmux current-workspace
{
  "id": "ws-15d62d15-a713-4805-a57f-88d415086baa",
  "title": "accel-144",
  "cwd": "C:\Users\Example\Work\project",
  "shell": "pwsh.exe",
  "surfaceId": "surf-931b3c87-a618-4309-9460-3a4d30ab99b7"
}
```

## Why

An agent running in a pane knows its own surface — `WMUX_SURFACE_ID` is in the env — but nothing maps that back to the workspace holding it. The workspace title is usually the only thing distinguishing two panes whose `cwd` and repo are identical, so anything an agent wants to attribute, route, or label per workspace currently has no name to use: an artifact it writes, a notification it posts, a scratch dir it keys off, or a `rename-workspace` / `close-workspace` call aimed at its own workspace.

The nearest existing command is `list-workspaces`, which carries `id` / `title` / `cwd`, and its `isActive` is exactly right for a UI query. It answers about the *focused* workspace, though, which is a different question from "the pane I'm in" once more than one pane is running. `list-surfaces` and `tree` are already caller-scoped, but their payloads carry no workspace id to join on, and `cwd` isn't a usable proxy — a workspace per branch in one repo puts several at the same path.

## How

This is wiring rather than new machinery — the resolution already exists. `resolveCallerTarget` (`src/main/v2-bridge.ts`) resolves the caller's `WMUX_SURFACE_ID` to `{ win, workspaceId }`, and `runBridge` already injects that id into any spec flagged `callerScoped`. Three touch points:

1. **`src/renderer/pipe-bridge.ts`** — `__wmux_currentWorkspace(workspaceId, surfaceId)` returning `{ id, title, cwd, shell, surfaceId }`.
2. **`src/main/v2-bridge.ts`** — a `workspace.current` spec, `callerScoped: true`, `requireResult: 'surface not found'`.
3. **`src/cli/wmux.ts`** — the `current-workspace` verb plus the `whoami` alias, `--surface <id>`, help entry and usage banner line.

## The one judgement call

When the surface can't be resolved — no `WMUX_SURFACE_ID`, or a stale one from a closed pane — `resolveCallerTarget` falls back to the focused window with no workspace opinion. So no id is injected, the renderer returns `null`, and `requireResult` turns that into an error with a non-zero exit.

Returning the focused workspace there would be perfectly plausible, but a caller couldn't then tell "I don't know" from "it's this one", which is the case this command exists to answer. I've erred toward the explicit miss — happy to flip it to the focused workspace if you'd rather it always return something.

`--surface <id>` asks on another pane's behalf, for symmetry with the other surface-taking commands. `--remote` works unchanged, since it rides the same V2 pipe.

## Verification

- `npx vitest run` — **1552 passed / 115 files** (6 new tests in `tests/unit/current-workspace.test.ts`, no Electron needed).
- `npm run build:main` clean; `npm run lint` shows no new findings in the touched files.
- Covered by the new tests: a caller in a **non-focused** workspace resolves to its own workspace; an unknown surface returns the explicit miss rather than the focused workspace; `--surface` targets another pane, including one in a second window.

Branched off `master` at v1.12.1. No version bump.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -343,7 +343,7 @@ The pipe server in `index.ts` handles V2 JSON-RPC methods. Most delegate to the 
 
 **Fully implemented V2 methods:**
 - `system.identify`, `system.capabilities`, `system.tree`
-- `workspace.create`, `workspace.close`, `workspace.select`, `workspace.rename`, `workspace.list`
+- `workspace.create`, `workspace.close`, `workspace.select`, `workspace.rename`, `workspace.list`, `workspace.current`
 - `pane.split`, `pane.close`, `pane.focus`, `pane.zoom`, `pane.list`
 - `surface.create`, `surface.close`, `surface.focus`, `surface.rename`, `surface.list`
 - `surface.send_text`, `surface.send_key`, `surface.read_text`, `surface.trigger_flash`
@@ -396,6 +396,10 @@ wmux new-window | list-windows | focus-window <id>
 # Workspaces
 wmux new-workspace [--title T] [--shell S] [--cwd D]   # --shell accepts args: --shell "ssh user@host"
 wmux close-workspace | select-workspace | rename-workspace | list-workspaces
+wmux current-workspace [--surface <id>]                # alias: whoami — the caller's OWN
+                                       # workspace {id,title,cwd,shell,surfaceId}.
+                                       # Explicit error on an unknown surface, rather than the
+                                       # focused workspace `list-workspaces` reports as active
 wmux ssh [ssh options] <user@host> [--title T]         # remote terminal in a new workspace (issue #78)
 
 # Remote wmux management (issue #78): drive another machine's wmux over an SSH tunnel
--- a/README.md
+++ b/README.md
@@ -311,6 +311,7 @@ wmux ping                          # Check if wmux is running
 wmux notify "Build complete"       # Send a notification
 wmux new-workspace --title "API"   # Create a workspace
 wmux list-workspaces               # List all workspaces
+wmux current-workspace             # The workspace THIS pane is in (alias: whoami)
 wmux ssh user@host                 # Remote terminal (OpenSSH) in a new workspace
 wmux ssh -p 2222 user@host --title "prod"  # Extra args are passed through to ssh
 wmux new-window                    # Second wmux window (e.g. for another monitor)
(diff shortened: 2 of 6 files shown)


amirlehmam:
Merged. Shipping in **v1.13.0**. This is a well-made PR — thank you.

**Your judgement call is the right one, and I'd have pushed back if you'd gone the other way.** Returning the focused workspace on an unresolvable surface would reintroduce exactly the confusion issue #143 existed to fix: window ≠ workspace, and "the pane I'm in" is not "the pane you're looking at". An agent in a background pane that silently received the focused workspace's title would mislabel its artifacts with no way to detect it had happened. The explicit miss with a non-zero exit is the answer a caller can actually branch on. Keeping it.

The framing in the PR body is also correct: this is wiring, not new machinery. `resolveCallerTarget` already did the hard part, `callerScoped` already injected the id, and the three touch points are exactly the three that were missing. The `--surface` flag for symmetry with the other surface-taking commands is a nice touch, and I confirmed it composes properly — `sendV2` auto-fills `caller` from `WMUX_SURFACE_ID` when the flag is absent, so both paths land on the same resolution.

Tests: 1552 passing on the merge with master, typecheck clean.

---

**One thing I had to fix on merge**, and it's a non-obvious repo trap rather than anything wrong with your reasoning: `resources/cli/wmux.js` is a **checked-in build artifact**, not a generated one. It's the CLI wmux puts on PATH for the terminals it spawns, and the copy that ships inside the package. Because your PR touched `src/cli/wmux.ts` only, `current-workspace` worked perfectly under `npm run dev` and would have been **absent from every install**.

That's the same shape as issue #81, where `wmux-hook.js` missing from `extraResources` broke every hook feature in packaged builds while dev was fine. The regeneration step is:

```bash
npm run build:main && cp dist/cli/wmux.js dist/cli/wmux-hook.js resources/cli/
```

`npm run verify:resources` catches the drift and gates the release workflow, so the tag would have failed rather than shipping a half-present command — the guard worked, it just fires after merge. Fixed in `52f2411`. Worth knowing for any future CLI change.

amirlehmam:
Released: **[v1.13.0](https://github.com/amirlehmam/wmux/releases/tag/v1.13.0)**.

Verified in the shipped artifact rather than just the source — `resources/cli/wmux.js` at the tag carries `current-workspace`, and CI's resource-drift gate passed.

Issue #128, opened 2026-07-29
grow wmux grow ;)


arnonuem:
Since wmux is already super close to herdr (https://www.youtube.com/watch?v=qnIu-Xu64H0), it is maybe a good idea to get some more ideas and inspirations from there (we did the autobackup recently). The code is Apache 2.0 and i guess worth looking at. What do you guys think would be awesome features wmux should grow on?

arnonuem:
Concrete inspiration from herdr's my coding tool integration — I went through the plugin source, and the wire protocol is small enough to copy in an afternoon.

The one idea worth stealing: the entire value sits in a single signal — "this agent is parked on a human." for example my coding tool emits it from one process-wide choke-point that every interactive wait already flows through: shell-command  approval, file-permission approval, questions, every menu. No per-prompt special-casing, nothing for the multiplexer to guess from screen content.

State model is two facts, three states:

```python
if awaiting_human:   return "blocked"
if run_depth > 0:    return "working"   # refcount, not bool — sub-agents
return "idle"
```
Four endpoints, newline-delimited JSON over a socket, one connection per message:



```
┌───────────────────────────┬────────────────────────────────────────────────────────────────────────────────┐
│ method                    │ why it matters                                                                 │
├───────────────────────────┼────────────────────────────────────────────────────────────────────────────────┤
│ pane.report_agent         │ the whole point — see at a glance which of 10 sessions needs you                  │
│ pane.report_agent_session │ ties a pane to a resumable session file, not a dead PID                        │
│ pane.report_metadata      │ model / context% / tokens, with a TTL so a crashed process doesn't lie forever │
│ pane.release_agent        │ agent actively let go → fall back to unknown instead of a ghost working        │
└───────────────────────────┴────────────────────────────────────────────────────────────────────────────────┘
```

Activation is three env vars the multiplexer injects into the pane (*_ENV=1, *_SOCKET_PATH, *_PANE_ID). Not set → the plugin doesn't even start a thread. Zero cost outside a pane, no install step.

Two things not to skip: a monotonic seq the receiver dedupes on (seq <= last_seq → drop), and an ack requirement. Without them a retry replays a stale state, or a lost idle strands the pane on working forever.

Minimum viable: report_agent + release_agent. Everything else is upside.

[...]

PR #206, opened 2026-08-26
feat(hub): agent office easter egg

kevmtt:
This adds a hidden easter egg: an "agent office" overlay that renders every agent in the window as a small pixel character working in a tiny office. The idea (inspired by the pixel-agents VS Code extension) is that the office doubles as a fun hub: you see at a glance which agent needs you, and you can watch the rest work. Every workspace gets a desk table with its name on it. Characters walk in through the door, sit down and type while working, stop and show a pulsing "!" speech bubble when blocked on you, take a coffee break when idle, and walk out when their pane closes. When an agent goes idle right as a colleague in the same workspace starts working, the idler walks over for a short chat first (a purely cosmetic "handoff" - wmux has no real signal for that).

Because it is mostly for fun, it ships hidden: no CLI command, nothing visible by default. A toggle at the bottom of Settings, General ("Easter eggs") enables it; only then does a pixel-invader button appear in the titlebar next to the settings gear, and Ctrl+Shift+O starts working.

It is not only decoration though:

- hovering a character shows its model, token usage, context percentage and how long it has been waiting (declared metadata, issue #128)
- clicking a blocked character opens its declared answer choices as buttons, relayed through the existing pane.answer_agent back-channel; answering never clears blocked, the agent confirms
- clicking a working character jumps to its pane; hovering a table highlights it and clicking switches to that workspace
- mouse wheel zooms (pixel scale 2 to 6, cursor-anchored), dragging pans when the office outgrows the window

Implementation notes:

- data comes exclusively from the existing rollupAgents / declared-state plumbing; no new state sources, no screen scraping
- the pure parts (layout generator, character state machines, camera math, sprite data) live in src/renderer/components/Hub/ and are unit tested (sprites, layout, sim, camera, rollup metadata passthrough)
- sprites are string pixel maps in TS rasterized once at mount: no binary assets, no release-process changes, all artwork original
- one Canvas 2D, fixed-timestep sim at 10 Hz with interpolated rendering at display rate
- closed means unmounted: no timers, no rAF, zero cost while the overlay is not open
- desks wrap at 6 per table row; up to 128 characters, beyond that a "+N" sign near the door
- design spec and plan are in docs/superpowers/

Screenshots

(dragged in below: office overview, hover stats, answer popover, table highlight, zoom, settings toggle, titlebar button)


kevmtt:
## The agent office
Full hub
hub-v2b

Hover details
02-hover-stats

Direct answer (this might need some more work)
03-answer-popover


## Settings to enable it
06-settings-toggle

## Button to show the hub
07-titlebar-button

docs/superpowers/plans/2026-08-26-agent-office-hub.md
kevmtt, line 1: I thought committing these files might be usefull for review and extending it further. We should delete it before merge though

amirlehmam:
Merged — this is sick. Shipping in **v2.2.0**.

The thing that made it an easy yes is that it adds no new machinery: it reads `rollupAgents` and the declared-state plumbing from #128, and answers blocked panes through the same `pane.answer_agent` back-channel `WorkspaceRow` already calls. No screen scraping, no new IPC surface, no state source that can drift from the sidebar's. And because the sprites are string pixel maps in TS rasterized at mount, there are no binary assets — the release process needed zero changes, which for this repo's packaging is worth more than it sounds.

`{hubOpen && <HubView/>}` with every timer and rAF cleaned up on unmount is the right call too: closed genuinely costs nothing, which is what lets an easter egg ship in the default build without anyone paying for it.

Verified on the merge result (your branch + master) before merging:

- `tsc --noEmit` clean
- 2191 unit tests pass, +44 from this PR (sprites, layout, sim, camera, rollup metadata passthrough)
- `eslint` clean on `src/renderer/components/Hub/`
- `verify:resources` clean

Two notes, neither blocking:

- Only `en.ts` carries the new keys, so the other 17 locales fall back to English. That's the documented behaviour and the fallback chain is tested, so nothing renders blank — just flagging it for whoever does the next translation pass.
- The `openHub` default of `Ctrl+Shift+O` is inert until the toggle is on, which is the right shape for an easter egg. Worth remembering it's now taken if something else wants that combo.

Thanks for the care in the follow-up commits — the pointer-capture, focus-stealing and answer-guard fixes are exactly the parts that would have been reported as bugs later.

Release: https://github.com/amirlehmam/wmux/releases/tag/v2.2.0

PR #213, opened 2026-08-31
feat(explorer): hidden-file visibility persists with the rest of the panel

derekneely:
The explorer's width and its per-root expansion both persist per workspace.
`showHidden` was a `useState(false)` inside `ExplorerPanel`, so it was the one
piece of panel state that reverted on a remount — a pane switch or a session
restore silently re-hid files the user had asked to see, while everything beside
it held.

It now lives in workspace metadata alongside `explorerExpanded`, read straight
off the workspace rather than mirrored into local state, so there is one value
and no resync effect.

**Per workspace, not global**, matching `explorerOpen` and `explorerWidth`: the
panel is a view onto one workspace's tree, so a repo where `.github/` matters
does not drag a repo where it is noise along with it.

**Absent stays falsy**, so sessions saved before this field come back hiding
hidden files, exactly as they did.

### The half that fails quietly

`replaceAllWorkspaces` names every field it carries and takes a `Partial`, so a
field the restore does not name is dropped with no error and no type complaint —
the shape of #145. The field is therefore declared in both `WorkspaceInfo` and
the `SavedSession` workspace entry and carried explicitly through all four
paths: `replaceAllWorkspaces`, the auto-save and the named-session save in
`App.tsx`, and `backupAutoSession` in `session-persistence.ts`. The round-trip
and the restore are tested rather than eyeballed.

The tree cache key already included the visibility flag (`treeCacheKey(root,
showHidden)`); a test pins that, since the value moved out of `useState` and the
cache identity it feeds must not follow it.

### Testing

- `tests/unit/explorer-show-hidden.test.ts` (new) — restore carries the field,
  absent stays `undefined`, per-workspace isolation, the toggle writes through
  `updateWorkspaceMetadata`, and the cache key still splits on visibility.
- `tests/unit/session-persistence.test.ts` — extended to round-trip the field
  and carry it through the version-change auto-backup.
- Full suite: 2592 passed / 166 files. `tsc --noEmit` clean on both projects.
- Manually: toggle hidden files on, switch panes, restart wmux — the toggle
  holds, and a second workspace keeps its own setting.

### DIFF
--- a/src/main/session-persistence.ts
+++ b/src/main/session-persistence.ts
@@ -37,6 +37,7 @@ export interface SessionData {
       explorerOpen?: boolean;
       explorerWidth?: number;
       explorerExpanded?: Record<string, string[]>;
+      explorerShowHidden?: boolean;
     }>;
   }>;
 }
@@ -131,6 +132,7 @@ function backupAutoSession(previousVersion: string): void {
         explorerOpen: w.explorerOpen,
         explorerWidth: w.explorerWidth,
         explorerExpanded: w.explorerExpanded,
+        explorerShowHidden: w.explorerShowHidden,
       })),
       sidebarWidth: windows[0]?.sidebarWidth ?? 260,
     };
--- a/src/renderer/App.tsx
+++ b/src/renderer/App.tsx
@@ -935,6 +935,7 @@ export default function App() {
             explorerOpen: ws.explorerOpen,
             explorerWidth: ws.explorerWidth,
             explorerExpanded: ws.explorerExpanded,
+            explorerShowHidden: ws.explorerShowHidden,
           })),
         }],
       };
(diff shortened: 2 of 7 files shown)


amirlehmam:
Merged, and shipped in 2.8.0. Thank you — this is a model PR.

The part I want to call out is the "half that fails quietly" section. You identified that `replaceAllWorkspaces` takes a `Partial` and names every field it carries, so an unnamed field is dropped with no error and no type complaint — and then carried the field explicitly through all four paths and *tested the restore* rather than eyeballing it. That is #145 exactly, and it's the failure mode that has bitten this file more than once. Getting there from a two-line `useState` fix is the difference between a patch and a fix.

The per-workspace scoping is also the right call for the reason you gave: the panel is a view onto one workspace's tree, so a repo where `.github/` matters shouldn't drag one where it's noise. And keeping absent falsy means sessions saved before this land exactly where they did.

Pinning `treeCacheKey(root, showHidden)` in a test was a good instinct — the value moved out of `useState` but the cache identity it feeds didn't, and nothing else would have caught that drift.

Squash-merged as `3294679`, with you credited.

Issue #53, opened 2026-06-21
Notify when an in-pane agent needs input (Claude Code Notification/Stop hooks, bell, wmux notify) — missing wiring


kevmtt:
## Use case

When I'm running an AI coding agent (Claude Code) inside a wmux terminal pane, I often step away while it works. I want wmux to **notify me whenever the agent needs my input** — for example when it pauses to ask a question, requests a permission, or finishes its turn and is waiting for me to respond.

Right now this simply doesn't happen. After investigating, the reason is **not** a misconfiguration on my side — there are several missing pieces in the wmux code. Nothing in the codebase ever turns an "agent needs you" event into a wmux notification.

This issue documents what wmux *does* today, why the "needs input" case slips through every existing path, and the concrete parts that need to change.

---

## How notifications work today

wmux can currently raise a notification through only **two** mechanisms:

1. **OSC escape sequences** — `src/renderer/hooks/useTerminal.ts:477-512` registers handlers for OSC `9` (iTerm2), `99` (kitty) and `777` (rxvt). If a program prints one of these, wmux fires a notification.
2. **Shell-integration "long command finished"** — `src/renderer/App.tsx:334-368`. The PowerShell/bash integration scripts report `report_shell_state` (`running` → `idle`/`interrupted`), and wmux notifies **only if the foreground command ran ≥ 5 seconds**.

The hooks that wmux installs into Claude Code are configured in `src/main/claude-context.ts:116` (`ensureClaudeHooks`) and are **`PostToolUse` only** (`trackedTools` at `claude-context.ts:157`). Those feed the sidebar/diff view via `hook.event` (`src/main/index.ts:1023`) — they are not notification triggers.

---

## Why "agent needs a response" never fires

Each existing path misses this specific case:

- **No `Notification` / `Stop` hook is installed.** Claude Code signals "I need permission/input" via its **`Notification`** hook and "I finished my turn" via its **`Stop`** hook. `ensureClaudeHooks` only writes `PostToolUse` entries, so neither event ever reaches wmux.
- **The terminal bell is swallowed.** Claude Code's default "I'm waiting" signal is the terminal bell (`\x07`). `useTerminal.ts` registers OSC handlers but **no `terminal.onBell` handler**, so the bell produces nothing.
[...]

Issue #187, opened 2026-08-22
OpenCode plugin: process.execPath is opencode.exe, not a JS runtime - all execFile calls fail silently


zutur:
## Problem

The auto-installed OpenCode plugin (`~/.config/opencode/plugin/wmux.js`) uses `process.execPath` to run `wmux.js`:

```js
const cli = process.env.WMUX_CLI;
const file = cli ? process.execPath : "wmux";
const argv = cli ? [cli, ...args] : args;
execFile(file, argv, { windowsHide: true }, () => {});
```

This assumes `process.execPath` is a Node or Bun binary that can execute JS files. In Claude Code that's true (`process.execPath` = `node.exe`), but in **OpenCode** it's the compiled binary:

```
C:\Users\stefan\.opencode\bin\opencode.exe
```

So every call becomes:

```
opencode.exe wmux.js agent-activity --surface ... --active
```

OpenCode doesn't recognize `wmux.js` as a subcommand and prints its help page. The `() => {}` callback swallows the error, so the failure is completely silent.

**Result:** The sidebar stays stuck on "Running" and never transitions to "Done", "Needs you", or any other state.

## Fix

Replace `process.execPath` with an actual Node binary when it doesn't point to a JS runtime. For example:

```js
function findNode() {
  const p = process.execPath;
  if (/\b(node|bun)(\.exe)?$/i.test(p)) return p;
  const fs = require("node:fs");
  const candidates = [
    process.env.LOCALAPPDATA + "\\Programs\\nodejs\\node.exe",
    "C:\\Program Files\\nodejs\\node.exe",
  ];
  for (const c of candidates) {
    try { if (fs.existsSync(c)) return c; } catch {}
  }
  return "node";
}
const NODE = findNode();
```

Then use `NODE` instead of `process.execPath`:

```js
const file = cli ? NODE : "wmux";
```

## Environment

- wmux: latest (plugin version 2)
- OpenCode: latest (`opencode.exe` compiled binary)
- OS: Windows 11
- Node: installed at `%LOCALAPPDATA%\Programs\nodejs\node.exe` but not on PATH inside OpenCode's process

zutur:
One thought — looking at this more closely, the plugin could potentially avoid the node dependency entirely by talking to the named pipe directly from within the plugin process. OpenCode's runtime supports `node:net`, so something like:

```js
const net = require("node:net");

function wmux(method, params) {
  try {
    const client = net.connect({ path: pipePath }, () => {
[...]

PR #215, opened 2026-09-02
fix(hooks): emit commands PowerShell can run, not 2>/dev/null

2more-ger:
## Summary

wmux writes Claude Code hooks as `node "…/wmux-hook.js" … 2>/dev/null || true`. That is bash. Grok — and anything else that honours `~/.claude/settings.json` — runs those commands in PowerShell on Windows, where `>/dev/null` is `Out-File` of `C:\dev\null`. Every hook then fails with exit 1 (`Could not find a part of the path 'C:\dev\null'`), so the sidebar never sees Grok activity.

This is not a one-machine quirk and not Grok-only. It hits any Windows agent that executes the injected commands outside bash. Claude Code itself often still uses Git Bash on Windows, which is why the wrapper survived.

## Why not skip rewriting existing hooks

`applyWmuxHooks` replacing wmux-owned entries on launch is how new events (`SessionStart` / `PreToolUse` / …) and `async: true` reach existing installs. Skipping the rewrite would preserve a local PowerShell patch until the next start — and would also freeze users on a stale hook set. User hooks are already left alone (issue #53). The durable fix is what we *write*, not whether we write.

A PowerShell-only spelling (`2>$null`) is equally wrong: it would break Claude Code under Git Bash. The helper already exits 0 when the pipe is absent, so the shell wrapper was never what kept a downed wmux from blocking the agent.

Bare `node "…"` is valid in bash, cmd and pwsh. Existing installs pick it up on the next wmux launch; no settings migration.

## Tests

- `tests/unit/claude-hooks.test.ts` — expected commands no longer carry the redirect; new assertion that no generated command contains `/dev/null`, `||`, or `2>$null`; legacy rewrite still upgrades an old `2>/dev/null` entry.
- `tests/unit/agent-integration.test.ts` — unchanged, still passes.

`npx vitest run tests/unit/claude-hooks.test.ts tests/unit/agent-integration.test.ts` — 36 passed.

### DIFF
--- a/src/main/claude-context.ts
+++ b/src/main/claude-context.ts
@@ -287,8 +287,17 @@ export function applyWmuxHooks(settings: any, hookScript: string): any {
 
   // PostToolUse passes the tool name as a positional arg; Notification/Stop
   // pass an --event flag so the helper reports an event type instead.
-  const makeToolCmd = (tool: string) => `node "${hookScript}" ${tool} 2>/dev/null || true`;
-  const makeEventCmd = (event: string) => `node "${hookScript}" --event ${event} 2>/dev/null || true`;
+  //
+  // Bare `node …` — not `2>/dev/null || true`. The helper already exits 0
+  // when the pipe is absent, so the shell wrapper never kept a downed wmux
+  // from blocking the agent. It DID break every consumer that is not bash:
+  // Grok (and anything else that honours ~/.claude/settings.json) runs these
+  // commands in PowerShell on Windows, where `>/dev/null` is Out-File of
+  // `C:\dev\null` and every hook fails with exit 1. Claude Code on Windows
+  // may still use Git Bash, so a PowerShell-only spelling (`2>$null`) is
+  // equally wrong. No redirect is valid in bash, cmd and pwsh.
+  const makeToolCmd = (tool: string) => `node "${hookScript}" ${tool}`;
+  const makeEventCmd = (event: string) => `node "${hookScript}" --event ${event}`;
 
   // The per-tool-call hooks run in the BACKGROUND. Every wmux hook is a pure
   // observer: it reports to the pipe and never blocks a tool or injects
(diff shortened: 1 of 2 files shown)


amirlehmam:
Merged, and shipping as **v2.9.1** (CI is building the tag now).

Thanks for this — the diff is small but the reasoning in the description is the part that made it easy to take. I verified all three claims before merging:

**1. The `|| true` really was cargo.** `src/cli/wmux-hook.ts` calls `process.exit(0)` on every path it can reach — the reply (`client.on('data')`), the deadline, and `client.on('error')`, which is exactly the "wmux not running" case. So the wrapper was never what kept a downed wmux from blocking an agent, as you say. Nothing depended on it.

**2. Legacy entries still get rewritten.** `stripWmux` keys on `HOOK_MARKER = 'wmux-hook'` — the script path, not the shell syntax — so an existing `2>/dev/null || true` entry is still recognised and replaced on the next launch. Your added assertion on the legacy-rewrite test pins that, which is the right place for it.

**3. The fix is complete, not partial.** `grep -rn "2>/dev/null" src/main/` now returns only your explanatory comment. `opencode-context.ts` and `kiro-context.ts` never emitted a shell wrapper, so there is no sibling emitter left spelling it the bash way.

Two things worth writing down for whoever reads this next:

- **You picked the right non-fix.** `2>$null` would have traded one broken host for another, and per-host detection is not available at write time — wmux writes `settings.json` once at launch and has no idea which agent, on which shell, will read it. "No redirect at all" is the only spelling that is correct for a file whose consumer is unknown. That is the durable property, and it is why this belongs in the emitter rather than in a Windows branch.
- **The one thing the wrapper did still cover** is `node` not being on `PATH` at all — that now surfaces as hook stderr instead of being swallowed. I am fine with it: in that state the hook was already reporting nothing, so the change is silence → a visible error, which is the better failure. (wmux resolves this properly for its own spawns via `WMUX_NODE`, #187, but a hook command is executed by the agent, not by wmux, so that resolution is not reachable from here. If it turns out to bite anyone, the fix is an absolute node path in the emitted command — a separate change, and not one this PR should have made.)

The observation that Claude Code on Windows often still uses Git Bash, which is why this survived this long, is the useful bit — it explains why the bug was invisible to the only agent wmux was tested against.

PR #217, opened 2026-09-05
perf: cut typing latency when many panes stream concurrently

kiromoussa:
## Problem

With several agent sessions streaming output at once, typing in any pane gets visibly laggy. Traced it to four compounding costs — none of them the keyboard path itself, which is already lean:

1. **One IPC message per ConPTY chunk.** Every `pty.onData` chunk was its own `webContents.send`, so N busy panes flood the single main→renderer channel that keystrokes also cross. Under load, echo queues behind other panes' output.
2. **`observePtyData` on every chunk of every surface**, on the main-process event loop that forwards keystrokes: ANSI-strip (two whole-chunk regex passes) plus up to 9 per-line handlers, ungated.
3. **`App.tsx` subscribed to the whole store** (bare `useStore()`), so any accepted store write re-rendered the entire tree — sidebar plus every workspace's split container. This is #141's shape one level up: the per-chunk paths stay out of the store, but each write that does land paid a full-App render.
4. **Every `WorkspaceRow` subscribed to the whole `surfaceProgress` object**, so one pane's OSC 9;4 tick re-rendered every sidebar row.

## Fix

- **`src/main/pty-data-batcher.ts` (new):** per-pane coalescing of PTY output before IPC. Leading edge is immediate — the first chunk after idle ships synchronously, so a lone keystroke echo never waits — then a 4ms trailing window batches the flood into one send per window. 256KB cap forces a mid-window delivery so a `cat bigfile` can't grow an unbounded string in main; `flush()` runs ahead of `PTY_EXIT` so trailing output still precedes the exit notification. Wired into both forwarders in `ipc-handlers.ts` (PTY_CREATE and `setupAgentPtyForwarding`). Sustained output settles at ~250 sends/s per pane instead of one per chunk.
- The observer now reads the batched bytes — identical lines, identical order, so parsing semantics are unchanged — which drops its regex cadence by the same factor.
- **`App.tsx`:** field-scoped `useShallow` selector over the 18 fields it actually reads. Actions are identity-stable, so App now re-renders only when workspaces/active/sidebar/shortcuts/notifications move.
- **`WorkspaceRow.tsx`:** each row selects only its own workspace's progress entries, shallow-compared; entries are identity-stable while unchanged, so other workspaces' progress no longer touches the row.

## Testing

- `tests/unit/pty-data-batcher.test.ts` (new, 8 cases, fake timers): immediate leading edge, window coalescing, sustained batching across windows, return-to-idle, cap-forced delivery, `flush()` semantics, byte-order preservation, `dispose()`.
- `npm run typecheck` clean; `npm test` 2708 passed (the one failure on my machine is the pre-existing `resolveExistingShellPath` pwsh test — this box has no PowerShell 7; it fails identically on a clean checkout of master).

### DIFF
--- a/src/main/ipc-handlers.ts
+++ b/src/main/ipc-handlers.ts
@@ -3,6 +3,7 @@ import * as path from 'path';
 import { promises as fsp } from 'fs';
 import { IPC_CHANNELS, SurfaceId, WindowId, WorkspaceId, AgentId, type InsertionResult, type ExplorerListError } from '../shared/types';
 import { observePtyData, clearActivity } from './claude-observer';
+import { createPtyDataBatcher } from './pty-data-batcher';
 import { clearAgentState, noteHumanInput, listAgentStates } from './agent-state';
 import { PtyManager } from './pty-manager';
 import { PtyLedger, reapOrphans } from './pty-ledger';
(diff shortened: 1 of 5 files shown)


amirlehmam:
Merged — shipping in **v2.9.3**.

Good diagnosis. The part I like most is that you didn't touch the keyboard path, which is already lean: all four costs are things that *crowd* it, and naming the main→renderer channel as the shared resource is the correct framing. `pty-data-batcher.ts` reads well and the leading-edge-then-trailing-window shape is the right one — the immediate first chunk is what keeps this from being a latency regression for the lone-keystroke case it's meant to fix.

Two details in there that are load-bearing and correctly handled, worth flagging so nobody "simplifies" them later:

- **Re-arming before delivering** in `closeWindow`. Without it, sustained output alternates immediate/batched on every expiry and you get half the sends you expected.
- **`flush()` ahead of `PTY_EXIT`**. Trailing output landing after the exit notification is a visible bug (the renderer paints "[process exited]" over bytes still in the window), and it's exactly the kind of ordering that only shows up under load.

The `useShallow` change on `App.tsx` is the one I scrutinised hardest, since a bare `useStore()` going field-scoped is the kind of change that silently drops a subscription. It reads correctly — actions are identity-stable so they cost nothing in the comparison, and the 18 fields match what App destructures.

One note for context, not a request: `npm test` here is **2716 passed**, including `resolveExistingShellPath` — that box just needs PowerShell 7. Nothing was wrong with your run.

Follow-up I'd merge if you're interested: `WINDOW_MS = 4` is currently a constant chosen by reasoning rather than measurement. Given you already have the harness to reproduce the many-streaming-panes case, a measurement of echo latency at 2/4/8ms under that load would either confirm 4 or find a better number — and either way it turns the constant into something with evidence behind it.

PR #218, opened 2026-09-05
fix(terminal): smooth touchpad scrolling + repaint after WebGL context loss

kiromoussa:
Two independent terminal-feel fixes, one commit each:

## 1. Touchpad scrolling is jerky (fractional-line accumulation)

Pixel-precision devices fire many wheel events per second with tiny sub-line `deltaY` values, and `wheelDeltaToLines`'s `Math.max(1, round)` forced every micro-event to a whole line — a gentle two-finger drag scrolled fast and jerky, line-at-a-time. Fix: a per-surface fractional-line accumulator carries the sub-line remainder between events and emits only whole lines, measured against the real cell height (host rect / rows) instead of a fixed 17px so the mapping tracks the user's font size. Notched wheels (`DOM_DELTA_LINE`/`PAGE`) keep the at-least-one-line-per-detent snap. The accumulator is cleaned up on surface reset alongside `surfaceMouseModes`.

## 2. Panes go blank after WebGL context loss

Chromium force-loses the oldest WebGL context past its ~16 cap — which happens exactly when panes are split/resized and several terminals flip visible at once. The existing `onContextLoss` handler correctly downgrades to the DOM renderer, but nothing repainted the buffer into it: the canvas the lost context left behind stayed blank until the next PTY write. This is the "text disappears when I resize/split panes" symptom. Fix: force a full `terminal.refresh(0, rows-1)` on the next frame once the DOM renderer is live.

## Testing

- `npm run typecheck` clean; `npm test` 2708 passed. (The one failure on my machine is the pre-existing `resolveExistingShellPath` pwsh test — no PowerShell 7 installed here; it fails identically on a clean master checkout.)
- Tested live: touchpad scrolling in a terminal pane, and split/resize churn past the WebGL context budget.

Independent of #217 (both apply cleanly to master in either order).

### DIFF
--- a/src/renderer/hooks/useTerminal.ts
+++ b/src/renderer/hooks/useTerminal.ts
@@ -428,13 +428,44 @@ function registerPromptMarks(terminal: Terminal, surfaceId: string | undefined) 
   return terminal.onScroll(() => handleAnchorScroll(terminal, surfaceId));
 }
 
-// Convert a wheel delta to a line count (sign preserved, magnitude ≥ 1).
-function wheelDeltaToLines(ev: WheelEvent, rows: number): number {
-  let amount: number;
-  if (ev.deltaMode === 1 /* DOM_DELTA_LINE */) amount = ev.deltaY;
-  else if (ev.deltaMode === 2 /* DOM_DELTA_PAGE */) amount = ev.deltaY * (rows || 24);
-  else amount = ev.deltaY / 17;
+// Per-surface fractional-line accumulator for pixel-precision devices
+// (touchpads, high-resolution mice). Those fire many events per second with
+// tiny sub-line deltaY values, and the old `Math.max(1, round)` forced EVERY
+// such micro-event to a whole line — so a gentle two-finger drag became a fast,
+// jerky line-at-a-time jump. We now carry the sub-line remainder between events
+// and only emit whole lines, which is what makes touchpad scrolling smooth.
+const wheelLineAccum = new Map<string, number>();
+
+// Round a line/page-mode delta away from zero so a notched wheel always moves at
+// least one line per detent.
+function snapWheelLines(amount: number): number {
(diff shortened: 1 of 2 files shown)


amirlehmam:
Merged — shipping in **v2.9.3**. Two genuinely independent fixes in one PR, one commit each, both correct; that's the right way to send this.

**Scrolling.** `Math.max(1, round)` forcing every sub-line micro-event to a whole line is a good catch, and the fix is better than just removing the clamp: keeping the snap for `DOM_DELTA_LINE`/`PAGE` preserves the detent feel on a notched wheel while the accumulator smooths pixel-precision devices. Measuring the cell height from the host rect instead of the fixed `17` is the part that will quietly help the most — that constant was wrong for every font size but the default.

**WebGL context loss.** This is the better of the two. The existing `onContextLoss` handler *looked* complete — it downgrades to the DOM renderer, which is the hard part — and the missing repaint is invisible until you notice the pane only comes back on the next PTY write. On an idle shell that's never, which is why this reads to users as "text disappears when I split panes" rather than as anything renderer-shaped. Pinning it to the ~16-context cap being hit exactly during split/resize churn is the detail that makes it diagnosable next time.

One thing I changed on merge: `wheelLineAccum.delete()` now also runs on the **PTY-exit path**, not only on `wmux:reset-terminal`. It's keyed by surfaceId and sits next to `surfaceMouseModes`, whose delete on that path carries a comment specifically about the map growing one entry per surface for the life of the process — same map shape, same problem. It's also right on its own terms: a fraction of a line owed to a gesture aimed at a now-dead process shouldn't be paid out to whatever opens on that surface next. Tiny, and entirely in the spirit of what you wrote.

For context: `npm test` here is **2716 passed** including `resolveExistingShellPath` — that box just needs PowerShell 7, nothing was wrong with your run.

Issue #212, opened 2026-08-31
Configurable defaults for a new workspace: terminal pane count and browser start URL


RMA-MUN:
### What are you trying to do?

I use wmux for multiple projects, and two startup behaviors force manual work every time I start a session:
1. New workspaces open with a hard-coded number of terminal panes. When I create a workspace, it opens exactly 3 terminal panes (T-shaped: 2 top + 1 bottom). Even when I want fewer or more, or a different arrangement, I have to re-split and re-arrange manually every time. wmux new-workspace via the CLI is inconsistent too — it opens a single pane instead.
2. The built-in browser panel opens blank. When I toggle the browser panel (Ctrl+Shift+I), it has no default page, and I have to type/click my usual starting URL each time.
I looked for settings for both and could not find them.

### What would you like wmux to do?

Add two configurable defaults, ideally in ~/.wmux/config.toml following the existing [terminal] / [remote] / [keys] pattern:
- Default terminal pane count / layout for a new workspace. A simple option to control how many terminal panes a fresh workspace starts with (and, if possible, how they are arranged), e.g.:
[workspace]
panes  = 3            # default terminal panes in a new workspace
layout = "grid"       # single | grid | left | down (optional)
This should apply consistently to the sidebar "+" button and wmux new-workspace (which currently opens a single pane), so the count is no longer hard-coded.
- Default browser start URL. A page the built-in browser panel opens to when it is first shown for a new workspace, e.g.:
[browser]
default-url = "http://localhost:3000"
Empty / unset should keep the current behavior (blank page). This is the start page, distinct from the existing default search engine.

### What do you do today instead?

- I press Ctrl+D / Ctrl+Shift+D and drag dividers to rebuild my layout after every new workspace.
- I navigate the browser panel to the URL I want each time.
- I can partially work around the pane-count gap with the Saved Layouts → set as Default feature, but that is a heavier flow and the CLI still opens a single pane regardless; there is no simple config value for "how many panes should a new workspace start with".

amirlehmam:
Both shipped in 2.8.0. You'd found a real inconsistency, and half of the second one turned out to be a bug rather than a missing feature.

## Pane count and layout

`[workspace]` in `~/.wmux/config.toml`, close to what you proposed:

```toml
[workspace]
panes  = 3         # 1-8
layout = "grid"    # grid | columns | rows | left | down | single
```

| `layout`  | |
|---|---|
[...]

PR #219, opened 2026-09-05
feat(browser): make dev-server auto-open a persisted pref, default off

kiromoussa:
## What

Auto-opening a detected dev server (Vite 5173, Next 3000, …) in the workspace's browser panel was a module-level runtime value, settable only via `~/.wmux/config.toml` `[browser] auto-open`, and **always on by default**. This promotes it to a persisted `BrowserPrefs` field (`autoOpenDevServer`) with a Settings → Browser toggle.

## ⚠️ Behavior change — default flips to OFF

A panel navigating to a website on its own — with no click — reads as the app doing something behind the user's back, which is a common complaint with auto-open features. With this change, detected ports are still tracked and shown on the workspace row; only the unprompted navigation is gated behind the opt-in. If you'd rather keep the old default, flipping `DEFAULT_BROWSER_PREFS.autoOpenDevServer` to `true` keeps everything else in this PR intact — happy to amend.

## Design

Follows the `defaultUrl` (#212) shape exactly: because Settings offers it too, it's a persisted pref rather than a module runtime value, so file-wins-at-startup and app-wins-at-runtime both fall out of writing to the same place. The `config.toml` key keeps working and now writes through `setBrowserPrefs`. `handlePortsUpdate` reads the pref at decision time via `useStore.getState()`.

New-field prefs migration is free per the `promptDefaultRev` note in CLAUDE.md (`{...DEFAULTS, ...stored}` fills in missing fields) — no rev bump needed since this field is new.

i18n: strings added to `en.ts`; the other locales fall back to the English default via `t(key, fallback)` until translated.

## Testing

`npm run typecheck` clean; `npm test` 2708 passed (the one failure on my machine is the pre-existing `resolveExistingShellPath` pwsh test — no PowerShell 7 on this box; fails identically on clean master). Toggled live: port detection still populates the workspace row with the pref off; navigation resumes when toggled on or set via config.toml.

Independent of #217 and #218.

### DIFF
--- a/src/renderer/App.tsx
+++ b/src/renderer/App.tsx
@@ -100,7 +100,6 @@ function findBottomPane(node: SplitNode): PaneId | null {
 // Effective runtime values — seeded from the built-in defaults, then widened/
 // toggled by ~/.wmux/config.toml at startup and on `wmux reload-config`.
 let activeDevPorts: number[] = DEFAULT_DEV_PORTS;
-let autoOpenDevPort = true;
 
 /**
  * Apply `~/.wmux/config.toml`'s `[browser]` section: dev-port detection + auto-open.
@@ -110,12 +109,14 @@ let autoOpenDevPort = true;
  */
 function applyUserConfigBrowser(state: any, browser: any): void {
   activeDevPorts = DEFAULT_DEV_PORTS;
-  autoOpenDevPort = true;
   if (!browser) return;
   if (Array.isArray(browser.devPorts) && browser.devPorts.length) {
     activeDevPorts = mergeDevPorts(DEFAULT_DEV_PORTS, browser.devPorts);
   }
-  if (typeof browser.autoOpen === 'boolean') autoOpenDevPort = browser.autoOpen;
+  // Like defaultUrl below, auto-open is a persisted PREF (Settings offers it too),
+  // so file-wins-at-startup and app-wins-at-runtime both fall out of writing it
+  // to the same place rather than to a module-level runtime value.
+  if (typeof browser.autoOpen === 'boolean') state.setBrowserPrefs({ autoOpenDevServer: browser.autoOpen });
   // The start page (#212) is a persisted PREF, not a module-level runtime value
   // like the two above, because Settings offers it too — so file-wins-at-startup
   // and app-wins-at-runtime both fall out of writing it to the same place.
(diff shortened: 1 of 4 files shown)


amirlehmam:
Merged — shipping in **v2.9.3**, with the default flipped back to ON. Details below, because you explicitly asked and the reasoning matters more than the one-word diff.

**The design is right and went in as written.** Promoting this from a module-level runtime value to a persisted pref is the correct call for exactly the reason you gave: because Settings offers it too, file-wins-at-startup and app-wins-at-runtime both fall out of writing to the same place. Following `defaultUrl` (#212) rather than inventing a second shape is what I'd have asked for. The `config.toml` write-through and reading the pref at decision time via `getState()` are both correct, and you were right that a new field needs no `promptDefaultRev` bump.

**The default, though, is the one thing I disagree with — and it's worth writing down why, because your reasoning isn't wrong, it's just answering a different question.**

"A panel navigating with no click reads as the app acting behind the user's back" is a good default-policy argument *for a new feature*. This isn't one. The behaviour predates the pref, so the default here isn't choosing a policy — it's choosing whether to **revoke** something every existing install already has. And the very mechanism you correctly identified as making migration free is what makes `false` expensive: because prefs persist as whole blocks and a new field is filled from `DEFAULTS`, defaulting off doesn't present anyone with a choice. It just switches auto-open off for every existing user on upgrade, silently, with no note attached.

There's already a precedent two fields up in the same interface. `openLinksExternally` documents it in one line: *"Default false — the panel is what wmux is for, and flipping the default would change behaviour for everyone on upgrade."* Same argument, same file.

So: the pref's job is to give the people it bothers a way out. That's a real complaint and you fixed it. It isn't a reason to take the behaviour from the people it doesn't bother, who never filed anything because nothing was wrong for them.

Merged as-is, then a follow-up commit flips `DEFAULT_BROWSER_PREFS.autoOpenDevServer` to `true`, corrects the Settings hint and `en.ts` string to match, and adds `tests/unit/browser-auto-open-default.test.ts`. The test exists because the value is a single word that reads like an oversight next to a comment arguing the opposite — precisely the shape of thing a later tidy-up flips back without noticing. It also pins the merge itself (a pre-#219 stored blob takes the default; an explicit stored `false` still wins), since the whole argument rests on that behaviour and not just on the constant.

Everything else — Settings UI, i18n, config write-through, the `getState()` read — is yours, unchanged. Nice, tight PR.

tonybolivar:
two llms speaking to each other bro

Issue #146, opened 2026-08-05
Plugin Support


arnonuem:
Unfortuantely we closed this issue https://github.com/amirlehmam/wmux/issues/112

Is it possible that plugins can be written for wmux?
I would love to write a plugin for myself to add the following behaviour to wmux:
- when ever CTRL+K is pressed, let it follow with the DEL key once.

amirlehmam:
Straight answer: I'm not building a plugin runtime, but your use case is shipped as a first-class feature in [0.44.0](https://github.com/amirlehmam/wmux/releases/tag/v0.44.0).

The reasoning on plugins: the thing you'd want a plugin for sits directly in the terminal input path. Third-party code there means an API I have to keep stable across releases, a sandbox story, and a new way for a bad plugin to eat your keystrokes or your PTY. That's a large permanent surface for what is, in your case, a one-line preference. So instead of a plugin API, wmux now lets you change what a key sends.

**What you asked for:**

```toml
# ~/.wmux/config.toml
[keys]
"ctrl+k" = "<C-k><Delete>"
```

Ctrl+K now transmits `0x0B` followed by the Delete sequence (`ESC [ 3 ~`), in that order, on every press — literally "Ctrl+K, then DEL once". Run `wmux reload-config` and it's live; no restart.

**The general form** is `"chord" = "sequence"`:

```toml
[keys]
"ctrl+alt+r"   = "clear<CR>"       # text outside <> is typed literally
"alt+left"     = "<Home>"
"ctrl+shift+q" = ""                # empty value swallows the key
```

Chords accept either `ctrl+shift+k` or the vim spelling `<C-S-k>`. Sequence tokens cover `<CR> <Esc> <Tab> <S-Tab> <BS> <Delete> <Ins> <Home> <End> <PgUp> <PgDn>`, arrows, `<F1>`–`<F12>`, `<C-x>`, `<A-x>`, `<Space>`, and `<lt>` for a literal `<`. Full table: [docs/config.md](https://github.com/amirlehmam/wmux/blob/master/docs/config.md#key-remaps).

**Worth knowing:**

- Remaps apply inside terminal panes and take priority over wmux's own shortcuts there — remap `ctrl+t` and Ctrl+T stops opening a tab while a terminal has focus. That's deliberate; it's the point of remapping.
- Modifiers match exactly, so a `ctrl+k` remap won't fire on Ctrl+Shift+K.
[...]

Issue #119, opened 2026-07-27
Shift+Enter inserts an unexpected newline


arnonuem:
Since one of the last updates shift+enter inserts a new line after the line break. What i get is this:

```
some message and shift enter[CR][LF]
[CR][LF]
another message and shift enter[CR][LF]
[CR][LF]
another message
```

I see this on windows.

amirlehmam:
Fixed in d38fec4, shipping in **v0.37.0**. Your `[CR][LF]` transcript was the whole diagnosis — you counted two where there should have been one, and that's exactly what was happening.

## What was going on

Shift+Enter is intercepted (added in #109, so "since one of the last updates" is right) to send `ESC`+`CR` instead of a plain `\r`, so TUI apps insert a newline rather than submitting:

```js
terminal.input('\x1b\r', true);
return false;
```

The `return false` looks like it stops there. It doesn't. Inside xterm 6:

```js
_keyDown(e) {
  this._keyDownHandled = false;
  if (this._customKeyEventHandler && this._customKeyEventHandler(e) === false) return false;
  …
}
```

Returning `false` makes `_keyDown` bail early — but it never calls `preventDefault`, and it leaves `_keyDownHandled` at `false`. So the DOM event is not cancelled, the browser goes on to fire `keypress`, and xterm's `_keyPress` (which checks `_keyDownHandled` to decide whether the key was already dealt with, and sees `false`) sends Enter's `charCode` 13 as a second, plain `\r`.

Your terminal was getting `ESC CR` **and** `CR`: one newline from us, one from xterm. Hence the blank line after every Shift+Enter.

## The fix

One line — `event.preventDefault()` before returning — since cancelling the keydown is what suppresses the `keypress` that produced the duplicate.

Small as it is, I moved the decision into `src/renderer/hooks/terminal-keys.ts` with tests on it, including one specifically asserting `preventDefault` was called. "Returning false is not enough" is the kind of invariant that reads as a redundant line and gets tidied away by the next person to touch it; now it fails a test instead.

Ctrl+Shift+Enter (zoom pane) is unaffected — it's still excluded, and there's a test for that too.

PR #222, opened 2026-09-07
fix(shortcuts): global shortcuts don't fire from a focused terminal

alejandronava562:
Two independent causes, one symptom. Together they account for `Ctrl+N`, `Ctrl+T`, `Ctrl+W`, `Ctrl+D`, `Ctrl+B`, `Ctrl+F`, `F1`, `Ctrl+2`–`Ctrl+8` and `Ctrl+Shift+[` / `Ctrl+Shift+]` doing nothing.

Closes #<issue>

---

## Cause 1 — xterm swallows the event before any listener sees it

Every global binding lives on a `document`-level, bubble-phase `keydown` listener:

| Listener | Location |
|---|---|
| Command palette | `App.tsx:608` |
| `ShortcutAction` table | `useKeyboardShortcuts.ts:448` |
| Numeric index rows | `useKeyboardShortcuts.ts:515` |
| Find / copy mode | `PaneWrapper.tsx:198` |

`CoreBrowserTerminal._keyDown` ends any key it handles with `this.cancel(event, true)` — `preventDefault()` **and `stopPropagation()`** — and it "handles" every bare `Ctrl+<letter>` by turning it into a control code (`Ctrl+N` → `\x0e`). The event dies at xterm's helper textarea, so none of those listeners run.

`Ctrl+Shift+P` was never affected: xterm produces no send-key result for it, doesn't cancel, and it bubbles normally. That asymmetry is what makes the bug look like "shortcuts only work after opening the command palette" — closing the palette moves focus off the terminal, and everything springs to life.

The clearest tell is the digit row: `Ctrl+1` and `Ctrl+9` work while `Ctrl+2`–`Ctrl+8` don't, because xterm maps only 2–8 to control codes.

`PaneWrapper.tsx:136` already documents this exact hazard for its own listener — *"Capture phase is load-bearing for keydown: xterm's hidden helper textarea is a descendant of this element and consumes the event"* — the global listeners just never got the same treatment.

### Fix

`useTerminal`'s `attachCustomKeyEventHandler` now asks whether wmux wants the keystroke and returns `false` when it does. From `@xterm/xterm@6.0.0`:

```js
_keyDown(e){
  if (..., this._customKeyEventHandler && !1===this._customKeyEventHandler(e)) return !1;
  // ... evaluateKeyboardEvent ... this.cancel(event, !0)
}
```

The `=== false` early-return is the **first** statement, before `evaluateKeyboardEvent` and before any `cancel()`. So returning `false` leaves the event alive and it bubbles to `document`, where the real handler runs and does its own `preventDefault` (which also suppresses the follow-on `keypress`).

Chosen over moving the listeners to capture phase because it touches one call site instead of four, and keeps the existing precedence order intact — `config.toml` remaps (#146), `Ctrl+C` copy, `Ctrl+V` paste and `Shift+Enter` (#119) are terminal-owned and still run first. The guard is placed last in the handler for that reason.

### One predicate, two callers

`isSafeToIntercept` moved from `useKeyboardShortcuts` into `utils/shortcut-binding` next to `bindingsEqual`, and `claimsKeyEvent` is built on it. Both the terminal (deciding whether to *release* a key) and the document listener (deciding whether to *act* on it) now ask the same function.

That equality is load-bearing. If the terminal released a key the document listener then declined, the keystroke would reach nothing at all — worse than the bug being fixed. Sharing the predicate makes that state unrepresentable rather than merely unlikely.

---

## Cause 2 — `Ctrl+Shift+[` / `]` could never match, focus aside

`nextSurface` and `prevSurface` cycle tabs within the focused pane. They have never fired, anywhere.

```ts
const eventKey   = e.key.length === 1 ? e.key.toLowerCase() : e.key;
const bindingKey = binding.key.length === 1 ? binding.key.toLowerCase() : binding.key;
```

`toLowerCase()` rescues a Shift-uppercased **letter** (`'N'` → `'n'`), which is what the comment above it was written for. It does nothing for punctuation, where Shift produces a *different character*: `]` → `}`, `[` → `{`.

`DEFAULT_SHORTCUTS` stores the unshifted form:

```ts
nextSurface: { key: ']', ctrl: true, shift: true },
prevSurface: { key: '[', ctrl: true, shift: true },
```

so the event arrives carrying `'}'` and the compare against `']'` cannot succeed. These two are the only affected bindings — every other Shift-carrying default is a letter, which `toLowerCase` already handles.

Worth noting `Ctrl+Alt+1`–`9` was immune throughout: `indexDigitFromEvent` reads `e.code` first (added for AZERTY), which incidentally protects the digit row from this entire class of bug.

### Fix

Fall back to the physical key, the same way `isLetterKey` already does in `hooks/terminal-keys`. `e.code` is stable under Shift, so `BracketRight` identifies the key whether it produced `]` or `}`.

The fallback is consulted **only** when the direct `e.key` compare has already failed *and* Shift is held. Non-US layouts are therefore untouched — there the recorder stored whatever `e.key` produced, and it matches directly. This exists purely to rescue the shipped defaults, which are written as the unshifted character but can only ever arrive shifted.

---

## Behaviour this changes

`SAFE_CTRL_KEYS` (`b d n t w f ,`) takes effect for the first time. Those keys have been reaching the shell all along *because of* Cause 1:

| Key | wmux action | Previously reached the shell as |
|---|---|---|
| `Ctrl+D` | split right | **EOF / logout** |
| `Ctrl+W` | close tab | delete-word-backward (readline) |
| `Ctrl+T` | new tab | transpose-chars (readline) |
| `Ctrl+B` | toggle sidebar | **tmux prefix** |
| `F1` | cheat sheet | help — htop, mc, vim, Far Manager |
| `F3` / `Shift+F3` | find next / previous | mc "View", htop "Search" |

The set is what the code always intended, so this PR does not change it — but tmux and WSL users will notice, and trimming it may be worth doing alongside. Happy to fold that in if you want it in the same change.

### Bare keys: named only

A bare key is released only when it is a **named** key (`F1`, `F3`, `PageUp`). Any single-character key still requires Ctrl or Alt.

`useKeyCapture` records a bare letter as a binding with no modifier required and no warning, so without that rule a rebind to `a` would swallow every `a` typed into the shell. A shortcut that doesn't fire from a focused pane is a papercut; a terminal that eats a letter is unusable. Nothing types `F3`, so named keys carry no equivalent risk.

`F1` is included deliberately. It opens the one screen that documents every other key, and a terminal is where you are standing when you need it, so leaving it to the PTY makes the cheat sheet undiscoverable in the default configuration. Everything here stays rebindable in Settings → Keyboard.

### Focus returns to the terminal when the cheat sheet closes

`ShortcutCheatSheet` focuses its own filter box on mount, and nothing handed focus back on close — the caret was left on `<body>`, so the next keystroke went nowhere and the pane had to be clicked again.

Latent until now: the overlay is bound to `F1`, which xterm swallowed, so it could only ever be opened from somewhere that was not a terminal. Making `F1` work is what surfaced it, so it is fixed here rather than left as a papercut this PR created the conditions for.

Refocuses the focused pane's active terminal through `surfaceTerminalRegistry`, on the next frame so the overlay has unmounted first. No-op when that surface isn't a terminal.

That covered the x, the backdrop and `Esc` — but not `F1` itself, which is the route a keyboard user actually takes. `F1` is a toggle, and the `wmux:toggle-cheatsheet` listener ran its own `setCheatSheetOpen(o => !o)` without going through `closeCheatSheet`, so the caret was still stranded after `F1` `F1`. Every close path except the most-used one was fixed. The listener now closes through `closeCheatSheet` and only opens directly; it reads `cheatSheetOpen` to know which way it is going, so the effect gains that and `closeCheatSheet` as deps. Re-registering one document listener when the focused pane changes is cheap.

`App.tsx` has no render harness — the single test importing it exercises a pure helper — so that path is hand-verified rather than covered. Called out rather than papered over.

> There is no focus-restore anywhere else in the codebase either — the command palette has the same gap, and it is the mechanism behind the "shortcuts only work after opening the command palette" symptom above (closing it parks focus on `<body>`, where the global listener finally sees the key). That one is pre-existing and not caused by this change, so it is left for a separate report.

## Tests

- `tests/unit/shortcut-binding.test.ts` — 28 cases: the combos that regressed, the shell keys that must still get through (`Ctrl+A/E/R/U/K/C/V`), bare modifier presses, bound vs unbound function keys, a bare printable key that is bound (must **not** be claimed), the digit row including AZERTY layouts and the `off` setting, and rebound shortcuts.
- `tests/unit/matches-binding.test.ts` (new) — 6 cases: both bracket combos matching their stored bindings, a layout that delivers the unshifted character directly, the two brackets not cross-matching, a bare `Ctrl+]` still not firing a Shift binding, and letter combos unaffected.

`npx tsc --noEmit` clean. 2779 passing.

Every user-facing claim above was also confirmed by hand on Windows 11: workspace and tab shortcuts firing from a focused terminal, `Ctrl+Shift+[` / `]` cycling tabs, `Ctrl+Alt+1`–`9` jumping to a tab, `F1` opening the cheat sheet and returning focus on close, and `Ctrl+C` still interrupting a running command.

> Note: rebased onto `v2.10.0` — replays with no conflicts; the tab-title commit and this work touch different parts of `useTerminal.ts`. Two suites (`pipe-client-latency`, `orchestration-status-vocab`) are load-sensitive and fail intermittently under a full parallel run on either base; they pass in isolation.

## Files

```
src/renderer/App.tsx                          33 ++++-
src/renderer/hooks/useKeyboardShortcuts.ts    56 +++++------
src/renderer/hooks/useTerminal.ts             26 ++++++
src/renderer/utils/shortcut-binding.ts        97 ++++++++++++++++++-
tests/unit/matches-binding.test.ts            56 +++++++++++
tests/unit/shortcut-binding.test.ts          145 ++++++++++++++++++++++++++++-
6 files changed, 378 insertions(+), 35 deletions(-)
```

Six commits, reviewable in order: the propagation fix, F1, the bare-printable-key guard, the shifted-punctuation match, the focus restore, then the `F1`-close path it missed.


amirlehmam:
Merged, thanks. The shared `claimsKeyEvent` predicate is the right shape — one answer for both the xterm handler and the document listener is what keeps a key from being released by one side and declined by the other. Also appreciated the F1 → `closeCheatSheet` focus-return detail; that would have been the next bug report.

Kept `SAFE_CTRL_KEYS` unchanged (they are the documented defaults) and noted the `Ctrl+D` trade-off in the release notes. Ships in v2.10.1.

PR #223, opened 2026-09-07
fix(shortcuts): stop a partial keyboard-pref write seating `undefined`

alejandronava562:
Fixes the renderer going blank ("wmux failed") with `Cannot read properties of undefined (reading 'ctrl')` after changing a keyboard setting, or on pressing `F1`.

Closes #<issue>

## The bug

`reconcileIndexModifiers` built its result with a spread:

```ts
const next = { ...prev, ...patch };
```

`applyIndexModifiers` always passes **both** keys, so the family the caller did not touch arrives as an explicit `undefined` — and the spread writes it over `prev`'s good value:

```
setKeyboardPrefs({ workspaceIndexModifiers: 'alt' })
  → patch = { workspace: 'alt', surface: undefined }
  → { ...prev, ...patch } → surface: undefined     ← clobbered
```

`MODIFIER_TRIPLE[undefined]` is `undefined`, and `matchIndexShortcut` reads `.ctrl` off it. That runs on **every keydown** via `handleIndexKey`, so the app doesn't fail one action — it dies through the root ErrorBoundary, unmounting the terminal panes with it:

```
TypeError: Cannot read properties of undefined (reading 'ctrl')
 > matchIndexShortcut  utils/index-shortcuts.ts:88
 > HTMLDocument.handleIndexKey  hooks/useKeyboardShortcuts.ts:533
```

`formatIndexShortcut` has the same shape, which is the `F1` path: the cheat sheet builds its "Select workspace 1–9" rows through it.

This is why the bad value never turns up in a bug report — `settings.json` and `localStorage` are both clean. It is manufactured on write, in memory, and only persisted afterwards.

## The fix

`patch.x ?? prev.x` instead of the spread. The `patch.x !== undefined` checks below still read the raw patch, so the collision-swap rule — the reason the function exists — is unchanged.

Two further layers, because the read sites should not be one bad value away from taking down the window:

- **`applyIndexModifiers` coerces both the patch and its base.** It is documented as "the one place a `KeyboardPrefs` value is produced" but coerced only on the load path, while `setKeyboardPrefs` passed caller-supplied input straight through — and the CLI and settings import both call it with unvalidated data. Coercing the base too means a store already holding a bad value heals on the next write instead of staying poisoned for the session.
- **`formatIndexShortcut` and `matchIndexShortcut` return `null`** for a value outside the union rather than throwing.

`formatBinding` likewise tolerates a missing entry — `shortcuts` is `DEFAULT_SHORTCUTS` spread with whatever was persisted, so a `null` there would throw on both the Settings screen and the cheat sheet.

## Repro

1. Settings → Keyboard → change either index-modifier dropdown.
2. Press any key. Window goes blank with "wmux failed".

Or press `F1` with a poisoned pref already in the store.

## Tests

- `tests/unit/keyboard-prefs.test.ts` (new) — the write path: unknown patch values, a base that is already poisoned, every legitimate value still passing through, and the collision swap still keying on which field the caller omitted.
- `tests/unit/index-shortcuts.test.ts` — the partial-patch clobber pinned directly, plus unknown values degrading to `null` in both readers.

`npx tsc --noEmit` clean.

> Note: `tests/unit/agent-browser-session.test.ts` and `tests/unit/agent-browser-teardown.test.ts` have 10 failures on pristine `origin/master` at 2.9.3, unrelated to this change.

## Files

```
src/renderer/components/Settings/KeyboardSettings.tsx   6 ++-
src/renderer/store/settings-slice.ts                   37 ++++++++++---
src/renderer/utils/index-shortcuts.ts                  28 +++++++++-
tests/unit/index-shortcuts.test.ts                     62 ++++++++++++++++++++++
tests/unit/keyboard-prefs.test.ts                      54 +++++++++++++++++++ (new)
5 files changed, 177 insertions(+), 10 deletions(-)
```

Two commits: the read-side hardening, then the write-side root cause. Independent of the terminal-shortcut PR — no shared files.


amirlehmam:
Merged, thanks. Textbook case of `{ ...prev, ...patch }` with an explicit `undefined` — and coercing at `applyIndexModifiers` rather than only on load is the right layer, since the CLI and the settings import both write through `setKeyboardPrefs` unvalidated. Ships in v2.10.1.

PR #224, opened 2026-09-07
fix(instance): finish what `WMUX_INSTANCE` promises

alejandronava562:
`src/shared/instance.ts` opens with:

> Set `WMUX_INSTANCE=<name>` to run wmux as a separate, side-by-side instance: the named pipe and APPDATA directory get a `-<name>` suffix, so a dev build can run alongside an installed production wmux without colliding.

Two things were not actually isolated, and both bite the **installed** app rather than the dev build — which is the wrong way round for a flag whose whole purpose is to protect the install.

## Bug 1 — the AppUserModelId is shared, so a dev build seizes the install's taskbar identity

`src/main/index.ts` hardcoded it:

```js
app.setAppUserModelId('com.wmux.app');
```

Windows resolves an AppUserModelId to exactly **one** executable. Starting a dev build repoints `com.wmux.app` at its own `node_modules/electron/dist/electron.exe`. From that moment:

- the **installed** app's taskbar button borrows the dev binary's identity — it reads **"Electron"**, with the Electron icon
- the user's **pinned** button is orphaned, a pin being keyed on that same id

Neither heals when the dev build exits. The association outlives the process, so the damage presents as a broken install long after the dev run is over — with no visible connection to the cause.

`build/installer.nsh` already documents the second half of this precisely:

> a pinned button's bitmap lives in the taskband store, keyed on the AppUserModelId (com.wmux.app), and nothing short of an Explorer restart or an unpin/repin refreshes it

That comment treats a stale bitmap as cosmetic and self-healing, which is fair for an in-place upgrade. It is not the same situation when the id has been repointed at a different binary: the name is wrong too, and re-pinning captures the *dev* executable, so the taskbar keeps launching the wrong wmux.

### Observed

On Windows 11, `npm run dev` from a clone, against an installed 2.10.0:

- taskbar group read `Electron - 2 running windows`
- "Pin to taskbar" produced a `.lnk` targeting `%TEMP%\wmux\node_modules\electron\dist\electron.exe`
- recovery took an unpin, an Explorer restart, and deleting 15 `iconcache*.db` files

Throughout, the install was verifiably intact — registry entry, `app.asar` hash, exe metadata, and the Start-menu shortcut's own AUMID all correct. Only the shell's id→exe mapping was wrong.

### Fix

`getAppUserModelId()` alongside `getPipePath()` and `getAppDataDir()`, carrying the same suffix:

```ts
export function getAppUserModelId(): string {
  return `com.wmux.app${suffix()}`;
}
```

Default is unchanged (`com.wmux.app`), so nothing moves for ordinary users. `WMUX_INSTANCE=dev` yields `com.wmux.app-dev` and the two stop competing.

## Bug 2 — shell integration resolved the pipe without the suffix

Already fixed in `fix(shell): honor WMUX_INSTANCE when resolving the named pipe` and included here because it is the same defect: an identity that ignored the suffix. A dev instance's PowerShell integration talked to the *production* pipe.

## Tests

`tests/unit/instance-identity.test.ts` — 5 cases: the bare default, the suffixed instance, pipe / APPDATA / AUMID suffixing in lockstep, and whitespace-only values ignored.

One is a guard rather than a regression test:

```ts
expect(getAppUserModelId()).toBe(builder.appId);
```

The installer registers shortcuts under electron-builder's `appId`. If the two ever drift, pins break for **ordinary users** rather than only for whoever is running a dev build — the same failure aimed at the wrong people. Better caught by a red test than by a bug report.

`npx tsc --noEmit` clean on both projects. 2779 passing.

## Files

```
src/main/index.ts                                        8 ++--
src/shared/instance.ts                                  19 +++++
src/shell-integration/wmux-powershell-integration.ps1   17 ++++-
tests/unit/instance-identity.test.ts                    53 +++++++++++
4 files changed, 93 insertions(+), 7 deletions(-)
```

Two commits, independent of the shortcut PRs — no shared files.

## Note

Found while running a patched dev build beside a real install to test PR-1. The flag did its job for the pipe and the session store, so the dev build looked safely isolated right up until the taskbar broke. Worth fixing precisely because the failure is silent, delayed, and points at the wrong culprit.


amirlehmam:
Merged, thanks. Good catch — the AUMID was the one identity `WMUX_INSTANCE` never suffixed, and the `getAppUserModelId() === electron-builder appId` guard test is the kind of thing that pays for itself the first time someone renames the product. The PowerShell pipe-name fix landed in both `src/shell-integration` and the `resources/` mirror, so the sync test is happy. Ships in v2.10.1.

PR #228, opened 2026-09-10
fix(browser): warn when agent wait drops per-call timeout

ConnorRobinson13:
## Summary

When `wmux browser wait <ref> <ms>` is routed through the agent-browser engine, the requested per-call timeout cannot currently be represented in agent-browser's `wait` CLI arguments. The timeout was silently dropped while the reference continued to be awaited using the existing global timeout behavior.

This change emits a warning so users know the timeout was not applied.

## What changed

- Warn when `browser.wait` has a non-empty ref and finite per-call timeout on the agent-browser engine.
- Preserve existing agent-browser argv behavior.
- Preserve child-process deadlines, stdout, and V2 responses.
- Keep web/CDP waits unchanged.
- Add focused tests for ref-plus-timeout, ref-only, timeout-only, zero, web-engine, non-wait, and non-finite cases.

## Scope

This PR intentionally does not change CI/CD, release workflows, CLI parsing, renderer code, settings, or agent-browser argv translation.

## Testing

Passed:

```text
npx vitest run tests/unit/agent-browser-wait-timeout-warning.test.ts tests/unit/agent-browser-verbs.test.ts tests/unit/cli-subcommands.test.ts tests/unit/agent-browser-routing.test.ts
4 test files passed
90 tests passed

npm run build:main
git diff --check
```

The full suite was also run. The new and related tests pass; remaining failures are existing Windows-specific/environment-dependent tests when run on Linux.


amirlehmam:
Merged as 0c4f6c5, shipping in **v2.10.3**. Thanks Connor — and thanks especially for the reasoning in that doc comment, which is the part I'd have had to reconstruct later.

You put the diagnostic in the right layer, for the right reason. `toAgentBrowserArgv` is pure and its tests deliberately *pin* that the timeout is dropped; a warning in there would have meant either coupling a pure function to output or relaxing a pin that exists on purpose. `runBrowserCommandForTarget` is where the resolved engine and the original params are both in hand, which is exactly what the check needs. `Number.isFinite(0)` keeping `timeout: 0` in scope rather than reading it as absent is the kind of detail that usually gets found by a bug report instead.

One follow-up on top (957cf80), which is about the *audience* rather than the logic:

`console.warn` in main reaches main's stdout. The caller is a different process — `wmux browser wait e1 5000` prints the JSON reply and nothing else — so the warning was visible under `npm run dev` and in no packaged build. The person who typed the ms couldn't see it, which is the person the PR is for.

So the check now returns the sentence instead of printing it, and the dispatch sends it to both places:

```
$ wmux browser wait e1 5000
{
  "ok": true,
  "warning": "browser.wait was given a 5000ms per-call timeout alongside ref \"e1\", but agent-browser's `wait` CLI has no per-call timeout flag — the timeout is dropped and the ref is still awaited under agent-browser's global default (AGENT_BROWSER_DEFAULT_TIMEOUT)."
}
```

`console.warn` stays for main's log (what I read after the fact). The `warning` field is attached only in this already-divergent case, so everywhere the two engines agree they still answer byte-identically — and it's added to a shaped payload, never substituted for one. Your seven tests all still pass unchanged; five more cover the reply channel.

The divergence note in CLAUDE.md now says it's *said* rather than silent, which is what it was missing.

PR #236, opened 2026-09-15
fix(sidebar): include detected agents in workspace status

### LINKED ISSUE #235: Workspace shows Idle while screen detection reports a working agent
### What happens

The workspace sidebar can show **Idle** while the agent roster banner reports working agents and terminal tabs show activity indicators.

Observed on Windows with Claude Code running on a Linux host through SSH and tmux. In the affected workspace, `wmux detect explain --surface <id>` returned `agent: "claude"`, `state: "working"`, and `ruleId: "claude.working.spinner"`, but the workspace label remained Idle. Two other tabs returned `unknown` / `no-rule-matched`. `wmux agent-state` returned empty states, blocked, and identified arrays, so there were no explicit agent-state reports.

The banner's working count is window-wide; the concrete contradiction is that a surface in this workspace was detected as working while its workspace said Idle.

Expected: workspace status should include screen-detected agent state, consistent with the roster. Missing or unmatched remote telemetry should not by itself imply that an identified agent is idle.

### Steps to reproduce

1. Open a terminal workspace in wmux 2.11.1 on Windows.
2. Connect through SSH to a Linux tmux session and run Claude Code without agent-state hooks reporting back to the local wmux instance.
3. Start a turn that displays a working spinner recognized by screen detection.
4. Confirm `wmux detect explain --surface <id>` reports `working`.
5. Compare the workspace's Idle label with the working indication in the roster/tab.

This was observed in an existing three-tab workspace; it has not been independently reproduced in a fresh profile.

### wmux version

2.11.1. The same implementation is present at master commit `ace738b92dd5db2a24971f781ba11f436075d3a6`.

### Windows version

Windows desktop; exact OS build not recorded.

### Anything else

Source inspection identifies inconsistent status inputs:

- [AgentRosterBanner.tsx lines 27–32](https://github.com/amirlehmam/wmux/blob/ace738b92dd5db2a24971f781ba11f436075d3a6/src/renderer/components/Sidebar/AgentRosterBanner.tsx#L27-L32) passes `agentDetections` into `rollupAgents`.
- [WorkspaceRow.tsx lines 189–201](https://github.com/amirlehmam/wmux/blob/ace738b92dd5db2a24971f781ba11f436075d3a6/src/renderer/components/Sidebar/WorkspaceRow.tsx#L189-L201) derives sessions from Claude activity, hooks, and declared states, without screen detections.
- [workspace-status.ts lines 119–137](https://github.com/amirlehmam/wmux/blob/ace738b92dd5db2a24971f781ba11f436075d3a6/src/renderer/components/Sidebar/workspace-status.ts#L119-L137) falls back to Idle when those inputs do not provide another status.

Suggested fix: use a shared agent-state rollup for the workspace label/dot and roster, retaining declared-state precedence and unknown-state semantics. A regression case should cover a workspace with no hook reports and one screen-detected working surface.

Related but distinct from #81: this concerns a workspace showing Idle despite an existing positive screen-detection result, rather than a running process being mistaken for active work.

homearanya:
## Summary

A workspace could show Idle while one of its remote terminal tabs was screen-detected as working. The roster consumed those detections, but the workspace row used only legacy activity and declared session state.

Use the existing roster rollup and workspace-state resolver for the workspace label, status colour and dot. Declared states still beat detection; blocked agents take priority over working agents; manual overrides still win. Identified agents with no known state show Unknown with neutral styling. Workspaces with no roster entries retain their existing shell/activity fallback.

Fixes #235.

## Validation

Checked on Linux with Node 22.19.0, using `npm ci --ignore-scripts`:

- Focused workspace status, mounted-row, roster and session tests: **86 passed**. Ten new mounted-row cases cover incoming detection updates without hooks, blocked/working mixtures, unknown states, declared-state precedence, workspace isolation and manual overrides.
- `npm run typecheck`: passed.
- `npm run build:main`, `npm run build:renderer`, `npm run verify:resources`: passed. Renderer build reports the existing large-chunk warning.
- `npm run lint`: 14 errors and 13 warnings, identical to unchanged base `ace738b92dd5db2a24971f781ba11f436075d3a6`; no diagnostics in changed files.
- `npm test -- --maxWorkers=4`: 2,745 passed, 19 failed, 53 skipped; 178 files passed and 11 failed. Running the same command on the unchanged base produced the exact same failing cases and suites, with 2,735 passing tests. Failures include Windows path/process assumptions and three suites unable to import the unbuilt native PTY module under this scripts-disabled Linux installation.

Windows Electron GUI verification remains outstanding. These Linux checks do not verify the installed Windows application or live SSH/tmux sessions.

### Windows follow-up

In an isolated test instance, run a remote agent without local state hooks and confirm a screen-detected working tab makes its workspace read Running with a running dot. Then verify a detected blocker reads Needs you, an identified unmatched agent reads Unknown, and declared states/manual overrides retain precedence. Compare the workspace with its entries in the roster.


amirlehmam:
Merged — thank you. The diagnosis is the valuable part here and it was spot on: the roster banner and the workspace row were answering one question from two different sets of facts, and pointing at the three exact source ranges in the issue meant there was nothing left to investigate. Sharing `rollupAgents` between them is the right fix and it is what shipped.

I did push a follow-up on top ([`fa74993`](https://github.com/amirlehmam/wmux/commit/fa74993)) that changes **where** the rolled-up state sits in the chain, and I want to explain it properly rather than just land it over you.

Your version ranks the rollup as its own priority level, above everything else. That breaks two things:

**1. It coarsens the line whenever the two agree.** The older signals answer a *finer* question than the rollup does. When a tool label is live, the row said "Reading file…"; with the rollup on top it says "Running…". The rollup is right, and it still throws away the better answer.

**2. `unknown` must fall through, not override.** This is the one that mattered. In the declared-state protocol (#128) `unknown` is not "no agent" and it is not "idle" — it means *"an agent is here and has not said what it is doing"*. Absence of a claim is not a claim, so it has to fall **through** to the heuristics rather than land on top of them. As its own priority level it did two visible things:

- printed the literal word **"Unknown"** on a row whose shell knew perfectly well it was running, and
- returned `''` for both the status class and the dot class, so the row lost its state dot entirely.

Your own test captured this as the intended behaviour, which is what made it easy to see the disagreement:

```ts
it('keeps an identified but unmatched agent unknown even with a running shell', () => {
  render({}, { shellState: 'running' });
  expectStatus('Unknown', '', '');   // ← a blank-dotted row saying "Unknown"
});
```

That case is common exactly where your bug lives: an agent identified by command, over ssh, with no hooks reaching this machine and no detection rule matching yet. Under your ordering every such pane goes to "Unknown" with no dot.

**What changed.** Instead of a new priority level, each roster count now *joins* the signal it is evidence for:

```ts
export function mergedAgentSignals(s: StatusTextInputs): MergedAgentSignals {
  const counts = s.agentCounts;
  return {
    blocked: Math.max(s.blockedSessions, counts?.blocked ?? 0),
    working: Math.max(s.workingSessions, counts?.working ?? 0),
    idle: s.claudeIsIdle || (counts?.idle ?? 0) > 0,
  };
}
```

The priority chain is then left byte for byte as it was. `unknown` contributes to no count and therefore changes nothing. Your bug is still fixed — a detected working agent with zero hook reports reaches priority 1.5 and the row says "Running…" — and a live tool label still wins over it, and `workspaceRow.unknown` goes away before it reached the other 20 locales.

Your `stateDotClassFor` extraction I kept as-is and I am glad you did it: that dot chain is the *third* thing that has to agree about one row, it had already drifted once, and it had no business living inline in the component. It is now documented as such.

Two small things I fixed while in there, both mine to catch, not yours:

- The multi-session summary is clamped to the sessions it can name. The roster can see agents a workspace has no tracked *session* for, so `Math.min(agent.working, s.sessionCount)` — otherwise "3/2 running" appears and reads as a bug rather than as detection working.
- Both chains were pushed over the repo's cognitive-complexity gate by the parallel chain; extracting `needsYouText`/`multiSessionText`/`shellStatusClass` brings them back under.

Your test file is kept and extended. Three cases were rewritten to assert the fall-through, and the invariant is now pinned as a **parity** rather than an expected string:

```ts
it.each(['running', 'interrupted', 'idle', undefined] as const)(
  'a silent agent renders identically to no agent at all (shell: %s)', ...
```

A silent agent must render byte-identically to no agent at all. Anything it changes, it changed by overriding a finer signal it knows nothing about — which is a stronger statement than any particular string, and it would have caught the "Unknown" case automatically.

Shipping in **2.11.2**, building now. Genuinely good report *and* patch — the source-range citations in #235 are the reason this took one session instead of three. Please keep them coming.

homearanya:
Thanks for fixing this @amirlehmam  It works great!

Issue #174, opened 2026-08-15
Crash-report requests should warn that a Windows minidump carries the process environment block


Ray0483:
Splitting this out of #150 at your suggestion.

### The problem

"wmux crashed, here's a dump" is a request maintainers make casually, and users comply casually. On Windows, a WER minidump includes the **process environment block**. User-scope environment variables are inherited by every process on the machine, so a dump from a developer's machine carries every secret that machine keeps in its environment — in cleartext, in a file the user is about to hand to a stranger and, more often than not, attach to a public issue.

This is not hypothetical. When I went looking through the eight dumps in #150 before uploading them, every one carried live credentials. The worst held eleven, including a production API key/secret/passphrase set and a bot token. I found that only because I stopped to check; my honest read is that if I'd been in a hurry I would have dragged the files into the attachment box and shipped the lot.

### Why this is worth a warning from wmux specifically

wmux is a terminal multiplexer. Its users are, by construction, developers who keep credentials in the environment so the shells wmux spawns can see them. That's not a misconfiguration — it's the normal setup for the exact population most likely to file a crash report here. The overlap between "people who will hit a crash" and "people whose environment is full of live keys" is close to total.

### Hand-scrubbing is not a safe workaround

Worth stating so it doesn't get suggested as the fix: rewriting byte ranges inside a minidump's `MemoryList` to blank out secrets risks producing a file that opens cleanly and analyses wrong. A subtly corrupted dump is worse than no dump, because it wastes the maintainer's time on a fake stack. If a dump can't be shared safely, the answer is not to sanitise it.

### Suggested remedies — any one of these would help

1. **Ask for the Event Log line first.** For a repeatable crash, `Faulting module name`, `Exception code`, `Fault offset` and the `Additional parameter` are often enough to fingerprint it and tell same-signature from new. They carry no secrets. In #150 that line alone turned out to be sufficient to answer the question the dump was originally wanted for.
2. **Warn at the point of asking.** Wherever the crash-report ask lives — issue template, docs, the message the app shows after a crash — one sentence: *a Windows minidump contains your environment variables; check before uploading, and say so if you'd rather not.*
3. **If a dump is genuinely needed, say what's in it.** Then the user can decide, or reproduce under a clean profile with a stripped environment and send that instead.

amirlehmam:
Shipped in **1.2.0** — and thank you for writing it, after I said twice I would and twice didn't.

[...]

PR #244, opened 2026-09-17
fix(updater): stop the portable-zip updater from tripping AV as a dropper

### LINKED ISSUE #3: Portable-zip auto-update triggers Norton AV lockdown — full system freeze, requires hard reset
## Description

Every time wmux self-updates (portable zip in-place update), Norton kicks off a deep scan/lockdown right as the new build relaunches. The whole system freezes — Task Manager won't open, can't even log out — and the only way out is a hard reset.

## Steps to reproduce

1. Run a portable-zip install of wmux with auto-update enabled.
2. Trigger an update (titlebar badge / Help button).
3. Let the update download, extract, and apply.
4. Norton's real-time protection (Insight / SONAR — "Norton Capture" scan) kicks in around relaunch.
5. System becomes unresponsive: Task Manager won't open, logoff doesn't work.
6. Only recovery is a hard reset.

## Root cause hypothesis

`applyStagedPortableUpdate` (`src/main/zip-updater.ts`) quits wmux, then hands off to a detached, windowless `cmd.exe` helper that:
1. waits for the old process to exit,
2. `robocopy`s the new build over the install root,
3. strips Mark-of-the-Web via `Unblock-File`,
4. relaunches `wmux.exe` via `start`,
5. (until now) deleted its own `.cmd` file.

wmux.exe ships unsigned (no code-signing cert — an accepted, documented tradeoff elsewhere in this repo). A hidden background script that silently overwrites an unsigned executable, immediately launches it, and then erases itself is a textbook dropper/self-cleanup signature — exactly the pattern AV behavioral engines (Norton SONAR / Download Insight in particular) are built to catch. That's the most plausible trigger for the scan/lockdown described above.

This can't be fully fixed without code-signing (a cost/business decision, not a code one), but the self-deleting-script part is a real, unforced red flag with no functional upside.

## Fix

See linked PR — drops the `del "%~f0"` self-delete from the generated helper script. Everything else about the apply step (unconditional relaunch on robocopy failure, MOTW strip, no caller-supplied paths embedded in the script) is unchanged.

Caveat: this removes one plausible trigger, it isn't a verified fix — no Norton install available to confirm the freeze stops. Worth a workaround note for anyone hitting this in the meantime: exclude the wmux install folder from Norton's Auto-Protect/SONAR before updating, or disable auto-update and swap the zip manually.

mkh63d:
## Summary

Fixes mkh63d/wmux#3 — Norton AV (SONAR/Insight) was locking up the whole
system on self-update. The portable-zip updater hands off to a detached,
hidden `cmd.exe` helper that overwrites the (unsigned) `wmux.exe` and
relaunches it, which is exactly the dropper/self-cleanup shape behavioral
AV engines are built to catch. This removes every trigger for that
signature and, along the way, fixes several correctness bugs the helper
had regardless of AV (quoting, error handling, cleanup).

## What changed

**AV-signature removal**
- Drop `del "%~f0"` — the helper no longer deletes itself after running
- Drop the hidden `Unblock-File` MOTW strip — it stripped nothing (robocopy
  already replaces the destination's zone-identifier stream) and was the
  strongest remaining signal
- The helper's console window is no longer hidden — `detached` +
  `CREATE_NO_WINDOW` doesn't actually hide a cmd.exe console on Windows, and
  a hidden detached script overwriting an unsigned exe is the dropper shape
  this PR is fixing. The window now shows a real title instead of a bare
  `findstr.exe` leftover.

**New leftover sweep** (since the helper no longer self-deletes)
- `sweepUpdateLeftovers` runs 60s after a packaged Windows start and removes
  stale `.cmd`/extract-dir/zip leftovers from `%TEMP%`, with conservative
  keep-rules on any doubt (recent, PID still alive, matching helper still
  fresh, symlink, etc.)

**Correctness fixes surfaced while doing this**
- Never quit into an update that can't actually run: the staged payload is
  checked before quitting, and the helper's `spawn` (not just `execFile`
  returning) gates the quit
- Fixed `cmd /c` argument handling for `%TEMP%` paths containing spaces
  (any "First Last" account) and paths containing `%VAR%`-style sequences
  cmd.exe would expand
- `robocopy`'s real exit code is now read correctly (`if errorlevel 8`,
  immune to an inherited `ERRORLEVEL` variable) — a failed copy no longer
  deletes the payload it needs to retry, and no longer claims a rollback
  that didn't happen
- Fixed two dialog-bypass bugs where a fast double-click on the update
  badge could quit wmux behind a still-open confirmation dialog
  (both the `ready` and `error`/retry paths)
- A retry now goes back through the confirmation dialog instead of
  silently reapplying; after a second failure it falls back to opening the
  release page
- The sweep and the general update-failure logging only ever record codes,
  never messages/paths (which can carry the Windows username)

**Docs**
- New design spec (`docs/2026-09-17-zip-updater-av-signals-design.md`)
  covering the AV analysis, verification results, and every review finding
- `CLAUDE.md` updated to describe the helper as it actually ships

## Testing

- `npm test` (added/extended coverage in `updater.test.ts`,
  `zip-updater.test.ts`, `use-update-fallback.test.ts` — 18 commits' worth
  of review rounds, each with its own regression test)
- Helper script run by hand end-to-end against a real `cmd.exe`: missing
  payload, no PID, forced robocopy failure, `%TEMP%` path with a space, and
  the happy path
- MOTW-strip removal verified: freshly downloaded payload carries no
  `:Zone.Identifier` stream after robocopy overwrite (A1 in the spec)

● User answered Claude's questions:
· Want me to open this PR now with `gh pr create` (mkh63d/wmux:master → amirlehmam/wmux:master), or just hand you the text to use via the compare link yourself? → Just give me the text

Here's the PR content — paste it into the compare page:

Title
fix(updater): stop the portable-zip updater from tripping AV as a dropper

Body
## Summary

Fixes mkh63d/wmux#3 — Norton AV (SONAR/Insight) was locking up the whole
system on self-update. The portable-zip updater hands off to a detached,
hidden `cmd.exe` helper that overwrites the (unsigned) `wmux.exe` and
relaunches it, which is exactly the dropper/self-cleanup shape behavioral
AV engines are built to catch. This removes every trigger for that
signature and, along the way, fixes several correctness bugs the helper
had regardless of AV (quoting, error handling, cleanup).

## What changed

**AV-signature removal**
- Drop `del "%~f0"` — the helper no longer deletes itself after running
- Drop the hidden `Unblock-File` MOTW strip — it stripped nothing (robocopy
  already replaces the destination's zone-identifier stream) and was the
  strongest remaining signal
- The helper's console window is no longer hidden — `detached` +
  `CREATE_NO_WINDOW` doesn't actually hide a cmd.exe console on Windows, and
  a hidden detached script overwriting an unsigned exe is the dropper shape
  this PR is fixing. The window now shows a real title instead of a bare
  `findstr.exe` leftover.

**New leftover sweep** (since the helper no longer self-deletes)
- `sweepUpdateLeftovers` runs 60s after a packaged Windows start and removes
  stale `.cmd`/extract-dir/zip leftovers from `%TEMP%`, with conservative
  keep-rules on any doubt (recent, PID still alive, matching helper still
  fresh, symlink, etc.)

**Correctness fixes surfaced while doing this**
- Never quit into an update that can't actually run: the staged payload is
  checked before quitting, and the helper's `spawn` (not just `execFile`
  returning) gates the quit
- Fixed `cmd /c` argument handling for `%TEMP%` paths containing spaces
  (any "First Last" account) and paths containing `%VAR%`-style sequences
  cmd.exe would expand
- `robocopy`'s real exit code is now read correctly (`if errorlevel 8`,
  immune to an inherited `ERRORLEVEL` variable) — a failed copy no longer
  deletes the payload it needs to retry, and no longer claims a rollback
  that didn't happen
- Fixed two dialog-bypass bugs where a fast double-click on the update
  badge could quit wmux behind a still-open confirmation dialog
  (both the `ready` and `error`/retry paths)
- A retry now goes back through the confirmation dialog instead of
  silently reapplying; after a second failure it falls back to opening the
  release page
- The sweep and the general update-failure logging only ever record codes,
  never messages/paths (which can carry the Windows username)

**Docs**
- New design spec (`docs/2026-09-17-zip-updater-av-signals-design.md`)
  covering the AV analysis, verification results, and every review finding
- `CLAUDE.md` updated to describe the helper as it actually ships

## Testing

- `npm test` (added/extended coverage in `updater.test.ts`,
  `zip-updater.test.ts`, `use-update-fallback.test.ts` — 18 commits' worth
  of review rounds, each with its own regression test)
- Helper script run by hand end-to-end against a real `cmd.exe`: missing
  payload, no PID, forced robocopy failure, `%TEMP%` path with a space, and
  the happy path
- MOTW-strip removal verified: freshly downloaded payload carries no
  `:Zone.Identifier` stream after robocopy overwrite (A1 in the spec)


amirlehmam:
Merged, and shipping in **2.13.0** (building now).

Thorough work — the AV analysis is the part I'd have got wrong on my own. `del "%~f0"` plus a hidden detached script overwriting an unsigned exe really is the textbook dropper shape, and the finding that the `Unblock-File` strip was stripping nothing (robocopy already replaces the destination's zone-identifier stream) is the kind of thing that only comes out of actually checking.

The correctness fixes that fell out are worth as much as the AV work: `if errorlevel 8` being immune to an inherited `ERRORLEVEL`, the two dialog-bypass races on the update badge, and especially *not quitting into an update that can't run* — that last one is the failure the user can't recover from, since wmux is already gone by the time it's discovered. Same instinct as the unconditional relaunch in the helper.

Also noted and appreciated: codes-never-messages in the sweep and failure logging. `%TEMP%` paths carry the Windows username, which is exactly why `wmux crash-report` refuses to read Event Log properties [10]/[11] (#174) — good that the updater now holds the same line.

`npm test` is green here at 3136 (198 files) with your suite in it.

